PR #34, opened 2018-07-14
Fix #32 #33

lfleck:
Fixes #32 #33 .

@scikit-multiflow/scikit-multiflow-repo-admins

PR #37, opened 2018-07-16
Apply Fix #33 also to DataStream 

### LINKED ISSUE #33: FileStream does not recognize int64 labels as classification problem
https://github.com/scikit-multiflow/scikit-multiflow/blob/105a1412be9b7df2cb82f4bfdaa046e38139f8c3/src/skmultiflow/data/file_stream.py#L206-L210

The if-statement evaluates to False for int64 labels (e.g. from sea_stream.csv). 
Possible patch could be `if np.issubdtype(self.y.dtype, np.integer):`

lfleck:
Extension to  #34 

Apply Fix #33 also to DataStream class


jacobmontiel:
Awesome, thanks for the previous and new fixes.

Issue #65, opened 2018-10-08
ARF not working on multi-class classification using LED, AGrawal, and Waveform


jche117:
### Expected behaviour
Adaptive random forest to work on multi-class classification
### Actual behaviour
Error during evaluate pre-quential 
### Steps to reproduce the behaviour

tree_size = [2,5,10,20,30,40,50]
rootFile = "/home/ubuntu/Documents/scikit-multiflow-master/Experiments/"
b = 500
    

stream = LEDGenerator()

stream.prepare_for_use()
    
for j in tree_size:
        #for b in batch_size:
    rf = AdaptiveRandomForest(n_estimators= j)
    filename = "LED_24features_" + str(j)+ "trees_"+ str(b) + "batch"
    print(filename)

    evaluator = EvaluatePrequential(show_plot = False, pretrain_size=b, max_samples = (j+50)*100 )
    evaluator.evaluate(stream = stream, model = rf)

*Note:* For comments and usage-related questions please consider the [users group](https://groups.google.com/forum/#!forum/scikit-multiflow-users).

jacobmontiel:
Confirmed. This is likely related to #63 which is being addressed. Will test once #63 in fixed.

jacobmontiel:
Fixed in #63 and #66

Issue #63, opened 2018-10-04
ARFBaseLearner incorrectly passes weights as classes (for HoeffdingTrees)


PGijsbers:
The `ARFBaseLearner`'s `partial_fit` calls its classifier's `partial_fit` using only positional arguments [here](https://github.com/scikit-multiflow/scikit-multiflow/blob/master/src/skmultiflow/meta/adaptive_random_forests.py#L429-L432):

```  
def partial_fit(self, X, y, weight, instances_seen):
    self.classifier.partial_fit(X, y, weight)

    if self.background_learner:
        self.background_learner.classifier.partial_fit(X, y, weight)
    ...
```
Weight is passed as third positional, however, for the default classifier which derives from `HoeffdingTree` this is wrong. The signature of [`HoeffdingTree.partial_fit`](https://github.com/scikit-multiflow/scikit-multiflow/blob/master/src/skmultiflow/trees/hoeffding_tree.py#L949):

`def partial_fit(self, X, y, classes=None, weight=None):`

The weight value passed is thus interpreted as `classes` and `weight` is None ([which then defaults to 1.0](https://github.com/scikit-multiflow/scikit-multiflow/blob/master/src/skmultiflow/trees/hoeffding_tree.py#L983-L984)).

If you are interested, I fixed this by using keyword arguments, while working on `predict_proba` mentioned in #62.
I only tested with the default ARFHoeffdingTree, but should every `partial_fit` function (that's a suitable `ARFBaseLearner`) take keyword arguments `weight` and `classes`? Because then it should generalize too.

PGijsbers:
Okay, so I noticed predictions change for the worse when making the change (0.653->0.469).
I'll go over it again to make sure I did not make any mistakes.
If everything is correct, I don't know that there is anything to do about it, but it makes sense to have a second set of eyes look at the changes?

jacobmontiel:
I will take a look into this. Just to confirm, the bug was affecting the weight assigned to the sample (posisson), which in the current version defaulted to 1.

PGijsbers:
Yes, from my understanding weights were -always- set to 1 because the passed weight was interpreted as `classes`. With the change, the Poisson RV is now the assigned weight instead.

PR #66, opened 2018-10-11
Arf predict proba

### LINKED ISSUE #62: Should `predict_proba` be implemented for AdaptiveRandomForest?
I would like to use a AdaptiveRandomForest model and predict class probabilities with it.
To this end, I made a `predict_proba` implementation for it. Is this something that is desired to be added to the main repository?
If so, I will add documentation/tests and create a PR.

PGijsbers:
## Changes
Fixes #62, #63, #65.

- Implementation of `predict_proba` for Adaptive Random Forest (#62).
- Fix a bug where weights were incorrectly passed (#63).
- Above bugfix solves #65 (technically only tested j=2,5,10 and rest still running).

### Remarks
In the discussion of #63 it was mentioned that the fix reduces performance.
However this was based on just the single unit test.
Using the same unit test, but changing the test set (by increasing the initial value of cnt), we see the following values:

| cnt  | master  | PR  |
|---|---|---|
|  0 | 0.6531  | 0.4694  |
|  1 | 0.5306 | 0.4694  |
|  2 | 0.4082  | 0.4694  |
|  3 | 0.5310  | 0.5306  |
|  4 | 0.3877  | 0.5510  |
|  5 | 0.5306  | 5714  |
|  6 | 0.5306  | 0.5918  |
|  7 | 0.4287  | 0.4898  |
|  8 | 0.408  | 0.5510  |
|  9 | 0.5510  | 0.5918  |

While we can do more thorough evaluations, for me it seems that the initial set-up just was an outlier, and there is no real performance decrease.
Further motivation is that if I increase the training set size on the PR version, performance does increase.
On the PR branch, using 150 samples for training results in a performance of 0.4694 (with cnt=0 determining the test set). Increasing this to 1500 increases performance to 0.5714, increasing the training set again to 15000 results in a performance of 0.7142.

## Checklist

- [x] Travis CI build passes.

### DIFF
--- a/src/skmultiflow/meta/adaptive_random_forests.py
+++ b/src/skmultiflow/meta/adaptive_random_forests.py
@@ -488,14 +488,14 @@ class ARFBaseLearner(BaseObject):
                     # (this effectively resets changes made to the object while it was still a bkg learner).
                     self.warning_detection.reset()
 
-        # Update the drift detection
-        self.drift_detection.add_element(int(not correctly_classifies))
+            # Update the drift detection
+            self.drift_detection.add_element(int(not correctly_classifies))
 
-        # Check if there was a change
-        if self.drift_detection.detected_change():
-            self.last_drift_on = instances_seen
-            self.nb_drifts_detected += 1
-            self.reset(instances_seen)
+            # Check if there was a change
+            if self.drift_detection.detected_change():
+                self.last_drift_on = instances_seen
+                self.nb_drifts_detected += 1
+                self.reset(instances_seen)
 
     def predict(self, X):
         return self.classifier.predict(X)
--- a/src/skmultiflow/meta/adaptive_random_forests.py
+++ b/src/skmultiflow/meta/adaptive_random_forests.py
@@ -164,6 +164,7 @@ class AdaptiveRandomForest(StreamModel):
         else:
             self.warning_detection_method = None
         self.instances_seen = 0
+        self.classes = None
         self._train_weight_seen_by_model = 0.0
         self.ensemble = None
         self._init_random_state = random_state
@@ -192,15 +193,18 @@ class AdaptiveRandomForest(StreamModel):
         raise NotImplementedError
     
     def partial_fit(self, X, y, classes=None, weight=1.0):
+        if self.classes is None and classes is not None:
+            self.classes = classes
+
         if y is not None:
             row_cnt, _ = get_dimensions(X)
             weight = check_weights(weight, expand_length=row_cnt)
             for i in range(row_cnt):
                 if weight[i] != 0.0:
                     self._train_weight_seen_by_model += weight[i]
-                    self._partial_fit(X[i], y[i], weight[i])
-        
-    def _partial_fit(self, X, y, weight):
+                    self._partial_fit(X[i], y[i], self.classes, weight[i])
+
+    def _partial_fit(self, X, y, classes=None, weight=1.0):
         self.instances_seen += 1
         
         if self.ensemble is None:
@@ -211,7 +215,10 @@ class AdaptiveRandomForest(StreamModel):
             self.ensemble[i].evaluator.add_result(y_predicted, y, weight)
             k = self.random_state.poisson(self.lambda_value)
             if k > 0:
-                self.ensemble[i].partial_fit(np.asarray([X]), np.asarray([y]), np.asarray([k]), self.instances_seen)
+                self.ensemble[i].partial_fit(np.asarray([X]), np.asarray([y]),
+                                             classes=classes,
+                                             weight=np.asarray([k]),
+                                             instances_seen=self.instances_seen)
     
     def predict(self, X):
         """Predicts the label of the X instance(s)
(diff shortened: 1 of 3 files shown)


PGijsbers:
There seems to be a difference between the predictions produced on the CI server and on my local machine. My first guess is that this is somehow because of the weights now being assigned correctly (as Poisson RV). However predictions are consistent locally across tests.  
Second option would be that some changes to the master branch might have changed the predictions. I will merge the up-to-date master into my development branch to check.

PGijsbers:
Merging the development branch into an up-to-date master branch locally also passes all tests.
As far as I can tell, the random state is controlled just fine (variable set in initialization and used when generating a weight). Any ideas?

jacobmontiel:
Hello @PGijsbers , some comments:

- Could you explain what do you mean by `cnt`?
- In `adaptive_random_forests.py` you import `check_is_fitted` but can't find where is being used?

PGijsbers:
With `cnt` I mean the local variable in the test function `test_adaptive_random_forests`. Changing that was just a quick and dirty way to test with different test sets.

I noticed that too in the PR, made the change locally but did not push yet because I was trying to figure out why the CI test generates different predictions.

PGijsbers:
Pushed the commit which removes the unused import. CI results should not change.

jacobmontiel:
Regarding the failed tests, the failure on different python versions is quite strange and unexpected. A similar behaviour was observed recently 21cb5de303b0f0f5bf71ef055d92f43534a95975. 

I will run a local test of your branch to confirm.

PGijsbers:
Good point. I do run Python 3.6 locally which does actually pass the test on Travis.

jacobmontiel:
Tests for ARF fail locally in Python 3.5.

```
...
>       assert np.alltrue(predictions == last_version_predictions)
E       assert False
E        +  where False = <function alltrue at 0x7f6f73baf7b8>([1, 1, 0, 1, 1, 0, ...] == [1, 1, 1, 1, 1, 0, ...]
E        +    where <function alltrue at 0x7f6f73baf7b8> = np.alltrue
...
tests/meta/test_adaptive_random_forests.py:42: AssertionError
...
>       assert np.alltrue(predictions == last_version_predictions)
E       assert False
E        +  where False = <function alltrue at 0x7f6f73baf7b8>(array([1, 1, ..., 1, 0, 0, 0]) == [1, 1, 0, 1, 1, 0, ...]
E        +    where <function alltrue at 0x7f6f73baf7b8> = np.alltrue
...
tests/meta/test_adaptive_random_forests.py:83: AssertionError
...
```


Also got failures on evaluators unit tests, did you merge the master branch? There are changes missing but

PGijsbers:
No I did not merge the master branch (as that should be done automatically), but I'll push a merge commit.
It seems that you encounter the same issue as Travis.

PGijsbers:
Merged latest version of master into the development branch. No merge conflicts, all tests pass locally with Python 3.6.

jacobmontiel:
Ingore my last comment about the evaluators, it was a local issue when I did the pull.

jacobmontiel:
I can confirm that tests work locally on Python 3.6.5.

I will also take a look into this. before making the decision.

PGijsbers:
I found a bug in my implementation which caused `predict_proba` to fail when called to predict for `X` with more than one sample.

The cause was the implementation which normalized the probabilities to 1. I now use scikit-learn's `preprocessing.normalize` instead (normalizing with l1 norm). This works both in the single and multi sample setting.

I added a unit test which calls `predict_proba` in this multi sample setting. It did fail with the old version, but passes the fix.

All tests pass locally (on Python 3.6).

jacobmontiel:
#### Regarding failing tests.
- The problem is unrelated to Travis, I can confirm that locally the same test fails on Python 3.4.5 and 3.5.5 while passing on 3.6.5 and vice-versa.
- The root cause is still unkown but seems (directly) unrelated to the code changes in the pull request.
- My suggestion is to create the test that passes on Travis with Python 3.4.x and 3.5.x and **temporary** disable the assert line for 3.6.5:
```python
import sys

if sys.version[2] != 6:
    assert ...
```

A new issue will be created before merging the PR for further investigation and to enable the test(s) for all Python versions.

Note: If you are using anaconda, then you can create a virtual environment with a different Python version to run the tests.

----
#### Other comments
During review I found a bug in the code unrelated to this PR/issue:

This section
https://github.com/scikit-multiflow/scikit-multiflow/blob/6ae5e9fad1f2b22b1735c355aacffdca20fa1fa5/src/skmultiflow/meta/adaptive_random_forests.py#L491-L498

should be inside of this `if` block:
https://github.com/scikit-multiflow/scikit-multiflow/blob/6ae5e9fad1f2b22b1735c355aacffdca20fa1fa5/src/skmultiflow/meta/adaptive_random_forests.py#L469

Could you add this fix to your PR?

Let me know if can be of any help.

PGijsbers:
Done

jacobmontiel:
Thanks @PGijsbers !

Can you disable the tests for the versions where they fail in order to get a pass from Travis? I will use the commit for the temporary deactivation in a new issue.

jacobmontiel:
Thanks @PGijsbers for your work, patience and your understanding on this change.

I have created #68 to further investigate the issue with the test.

This change is ready to merge with the master.

Issue #56, opened 2018-09-18
The memory contabilization might not take into account the objects pointed by the variables


smastelini:
Greetings,

I think I found an issue in the memory costs calculation of the Hoeffding Tree-based algorithms.

The use of `sys.getsizeof` alone over an class object, list, dictionary, among others, does not account for the actual sizes of the properties contained within the structures, but only the pointers to them. For example:

```python
import sys

sys.getsizeof([]) # returns 64
sys.getsizeof(1) # returns 28

sys.getsizeof([1]) # returns 72
sys.getsizeof([1, 1]) # returns 80
```

This simple example shows that Python calculates the byte size only considering the pointers to the actual referenced objects. 

This kind of issue is discussed in https://code.tutsplus.com/tutorials/understand-how-much-memory-your-python-objects-use--cms-25609.

An elegant solution is presented in https://goshippo.com/blog/measure-real-size-any-python-object/.

Applications which need to measure the memory costs of a solution would benefit from such strategy.

jacobmontiel:
Functionality is added in afdf7c7f209112bf7c931d01c174c97690fc023a

Issue #18, opened 2018-04-03
Include running time as evaluation metric


jmread:
The running time should be included as an evaluation metric. It is already being logged for the overall process, but it would be nice to get a plot over time. It could also be interesting to split into training time / testing time.

smastelini:
Hi @jmread, I intend to implement this capability and also the memory usage accounting soon.

@jacobmontiel and @jmread could you give me some guides or suggestions of how I can make those changes maintaining the structure and standards of the framework? 

Would adding new classes to the measure collection be the best option? In this case, these metrics would be common for all the prediction tasks. This could change the way how the framework makes the distinction among the prediction tasks unless the new metrics are added to all tasks (in `skmultiflow/utils/constants.py`). For example, when computing the following set operations: 

https://github.com/scikit-multiflow/scikit-multiflow/blob/fdf348712e39db3c6b79ed96669b9e623095cedf/src/skmultiflow/evaluation/base_evaluator.py#L203-L218

smastelini:
Hello again @jacobmontiel

When you mention evaluating `n` models, are you referring to ensemble predictors or the comparison of different techniques?

- In the case of ensembles, I think that might be interesting to report the total running time, excluding individual predictors, at least initially.

- Additionally, reporting the training and testing time separately (in addition to the total running time) could be very useful too.

jacobmontiel:
I meant *n* models being compared (either simple models or ensembles). Now, if we consider multiple time points to track then it may be worth to have a class or method for this specific purpose.

smastelini:
Hi @jacobmontiel,

I agree with you. If the running time and memory costs are going to be reported at the same stream "points" (or at the same frequency) as other metrics, e.g., accuracy, then it may be worth creating classes like `*TimeAndMemoryMeasurements` for these tasks. 

In fact, the major reason I see to do this, would be writing the measured values into the output log and plotting the measured results.

jacobmontiel:
I prefer to have separate trackers for `memory` and `time`. I think `time` has to be tracked by default (as currently is) and could be added to the plots in the form o a legend (text). On the other hand, `memory` could be optionally selected and plotted.

smastelini:
You are right. Computing time will add little extra processing to the framework. On the other hand, calculating the total memory used by the models is costly when done at possibly small time intervals. My bad!

jacobmontiel:
[...]

PR #67, opened 2018-10-11
Support for running time and model size accounting

smastelini:
Hi @scikit-multiflow/scikit-multiflow-repo-admins,

This PR is focused on issues #18 and #56.

I've created a RunningTimeMeasurements class which accounts for the training, testing and total running time of each evaluated predictor. To ignore the time spent on other operations I had to call the running accounting methods directly before the `partial_fit` and `predict` methods on `evaluate_prequential` and `evaluate_holdout`. For instance:

https://github.com/smastelini/scikit-multiflow/blob/05f07b67809807d2564f7ad04faa35b55d46f85c/src/skmultiflow/evaluation/evaluate_prequential.py#L277-L279

Do you have any other suggestions?

Regarding the model size accounting, a function was added in `skmultiflow/utils/utils.py`. This tool is only called when the metrics are updated in the `skmultiflow/evaluation/base_evaluator.py`. The default returning unit is MB.

In addition, I provide tests for the new functionalities. 

The visualizations for the new metrics were implemented as subplots without axes ticks and bounding boxes, where the measurements are annotated. That was the best solution I could find, considering the auto re-scaling of the elements that is performed periodically. Could a better option be implemented? I would happily implement the required changes.

In summary, the changes proposed in this pull request are:
- Running time is added as a new performance metric
- Model size is added as a new performance metric
- Automated tests and visualizations for the new features are also added

Checklist:

- [x] Travis CI build passes.
- [ ] Test Coverage is maintained (threshold 0.2%)

* An observation:
Some of the current existing commits in skmultiflow/MASTER might appear as authored by @jacobmontiel  and commited by me in this PR. This was an effect of the "git rebase" I performed in my commits to avoid polluting the timeline. I merged many minor incremental changes into few steps. Sorry for the possible mess...

Thanks in advance for the help!


jacobmontiel:
Hello @smastelini , could you add a screen-shot of the updated layout?

jacobmontiel:
Hello @smastelini 

Thanks for the image, it helps a lot for the following discussion. Using new subplots has the negative side effect that other plots are squeezed while the annotations subplots waste a lot of space.

First, from how the annotations look I think we can fit all the information in one line per model (plus the headers):

``` text
Model   Train  |  Predict  |  Total  |  Mem (MB)
M0     ##.##s      ##.##s    ##.##s      ##.##
```

Suggestions:
1. Numeric values should only display 2 decimals max.
2. In order to avoid using new subplots we can either: 
   a) Use the last subplot as anchor for time/mem or
   b) Use the figure object to add this annotation

Does this sounds reasonable?

smastelini:
Hi @jacobmontiel,

My only concerns of using annotation directly on the `fig` object are the plot boundaries. In case of multiple metrics or models being compared, the annotations could not be visible. But I think that in "normal usage" scenarios the annotations will perform just fine!

About fitting all information in one line, I agree completely! It is much more clean and compact. I will try to anchor the new annotations on the last subplot and use the suggestions that you gave me.

Thank you for the suggestions!

smastelini:
Hi @jacobmontiel,

I've implemented the suggestions. Following I present an updated plot's screenshot. 

screenshot from 2018-10-15 10-30-03

Is that what you had in mind?

jacobmontiel:
Yes, this looks really good. There is something to take into account, the number of columns in the figure legend is currently set to the number of models, see here

https://github.com/scikit-multiflow/scikit-multiflow/blob/ee035f998b845e0339e8bf7e951163b1e31ab6cd/src/skmultiflow/visualization/evaluation_visualizer.py#L314-L317

we will need to limit it, I would say to max 2 columns. Maybe even set it to the *lower right* while keeping the time/memory to the left. What do you think?

To confirm this you can run the evaluation example for multiple models in the [documentation](https://scikit-multiflow.github.io/scikit-multiflow/skmultiflow.evaluation.evaluate_prequential.html#module-skmultiflow.evaluation.evaluate_prequential).

Finally, I see that you are also updating the holdout method. We just need to make sure that the visualization works on both evaluators.

Let me know if I can be of any help.

smastelini:
Hi @jacobmontiel,

Sorry for the delay. I changed the legend position to "bottom right" and placed the annotations in the left, as requested. Following I present the obtained screenshots including the example you suggested.

Firstly, a single model and Prequential:

prequential_single

Next, using two models and the Prequential approach:

prequential_two

Using a single model and the Holdout strategy:

holdout_single

Two models and the Holdout strategy:

holdout_two

What do you think?

jacobmontiel:
Thanks @smastelini !

This is a great improvement for the overall visualization. Being picky I think we just need to increase a little the space between the x-axis label ("Samples") and the test/mem metrics to avoid clutter. This is also to show that those metrics apply to the whole figure and not only to the last sub_plot.

Is this reasonable?

Regarding the coverage report, it seems that `codecov/patch` is ignoring previous commits but `codecov/project` is ok. This change is almost ready.

smastelini:
Hi @jacobmontiel,

I increased the distance between the last subplot x axis and the annotation . I had to tune the bottom adjustment of the subplots a little bit to achieve the following result. 

screenshot from 2018-10-15 17-38-40

What do you think? This solution is acceptable? I think that there is space to comfortably fit one more model in the comparison. More than three predictors could lead to the annotations being placed out of figure's limits.

jacobmontiel:
From the examples I noticed that some methods do not support the size measurement (which is fine). A nice way to show this would be to change

https://github.com/scikit-multiflow/scikit-multiflow/blob/6df9567b03d809ccb05be395a19808e7adef3781/src/skmultiflow/visualization/evaluation_visualizer.py#L485-L488

to

```python
if constants.MODEL_SIZE in self.metrics:
    size = memory_time['model_size'][i]
    text = '{: ^19.2f}'.format(size) if size > 0.0 else '{: ^19}'.format('-.--')
    text_info += text
```

We could either use `'-.--'` or `'NA'` to make it clear that the measurement is not available (instead of 0.00). What do you think?

Finally, I am considering that time and memory should be displayed by default. Do you think that the overhead from this functionality is justifiable? I believe that the user can benefit a lot from this information but don't wan't to negatively impact the evaluation time for normal usage.

jacobmontiel:
I think that the last example looks good. The updated layout provides useful information without clutter. As you mention, a third model could fit but more than that might be a problem. However, that is not a "normal" use-case and is even discouraged in the evaluators documentation.

smastelini:
Hi @jacobmontiel,

I have to investigate it first, but I think that the obtained zeros could be a consequence of unit conversion. The `calculate_object_size` function I have added returns the results in MB as default. But other conversions are also supported, such as, kB and byte. 

Since the original byte size is divided by `2 ** 20`, it is possible that the obtained number get too close to zero and the string formatting `{: 16.2f}` ends to making it always this value. I will investigate it further. I would prefer solving the possible "lack of support" primarily, but if it is not possible, your suggestion will be more than welcome!

Regarding the time and memory accounting overhead, in my tests I could not perceive great changes in the running time. But I mainly evaluated small cases. Among the two metrics, the model size accounting is the most costly, for sure.

It is worth mentioning that in case those two metrics become a default in the framework, it might be interesting letting their report in the log file be optional. We may face issues when validating the generated logs in the `tests/evaluate_holdout.py` and `tests/evaluate_prequential.py` due to the nature of these metrics be totally hardware/architecture dependent. 

What do you think?

Thanks again for your help and support!

jacobmontiel:
Thanks for the feedback. Taking into consideration your comments then we can roll-out this version where time/mem are optional and decide in the near future if they should be used by default.

About the testing, it might get a little bit complicated but I don't think it is a real concern. However, it is nice to have this in mind if we decide to make them default options (and include them in the report).

smastelini:
I performed a test locally to assess whether the memory measurement was working at all. I changed the `calculate_object_size` function's return unit to byte and performed a simple test.

screenshot from 2018-10-16 10-56-05

I am not very familiar with NB in streaming contexts, but it seems that the memory cost of this method is very small. Does it make sense for you?

jacobmontiel:
Oh, I see now, it makes sense. What about using kB as a compromise for small and large models? I think there is enough horizontal space.

smastelini:
I agree with you suggestion! I added kB as the default returning model size unit. In the future, if we see the necessity to change it back to MB, or even to byte, the modifications are pretty straightforward.

jacobmontiel:
I think this PR is ready to be merged with the master.

Coverage delta is justified.

Thanks @smastelini !

PR #75, opened 2018-10-27
Fix 40

### LINKED ISSUE #40: Predictions shall be returned as numpy.array instead of lists
### Expected behaviour
As discussed in #39, Predictions from `predict()` and `predict_proba()` shall be returned as `numpy.array` instead of lists.

This is to be consistent with `scikit-learn` which is also the expected data type for many users.

### Actual behaviour
Currently, we are returning predictions as `list`.

Note: A fix is to cast predictions before returning them:

```python
# current
return predictions
# suggested
return numpy.array(predictions)
```

This change is to be propagated to all models for consistency.

PGijsbers:
Fixes #40.

- Ensure `predict` and `predict_proba` functions always return a `numpy.ndarray`.
- Add assert in existing unit tests to ensure this functionality.
- Update docstrings.

Checklist

- [x] Travis CI build passes.

### DIFF
--- a/src/skmultiflow/lazy/sam_knn.py
+++ b/src/skmultiflow/lazy/sam_knn.py
@@ -376,7 +376,7 @@ class SAMKNN(StreamModel):
         for i in range(r):
             distancesSTM = SAMKNN.get_distances(X[i], self._STMSamples)
             predictedLabel.append(self.predictFct(X[i], None, distancesSTM))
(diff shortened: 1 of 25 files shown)


jacobmontiel:
Awesome! Thanks for taking care of this @PGijsbers 

Changes are correct and tests are passing. I will merge this PR with the master.

PR #76, opened 2018-10-29
Fixiing and improving model size accounting

### LINKED ISSUE #71: model size mtric not working with HAT
When evaluating HAT with model size metric, the evaluator crashs with error ```'int' object has no attribute 'calc_byte_size_including_subtree' ``` At the first example after n_wait.
 
Part of the problem is children missing .values().
Fixing this makes `calculate_object_size(HAT)`raise the error at `sys.getsizeof(obj)` : an interger is needed.
So the problem is in `__sizeof__()`

---
[*Note:* For comments and usage-related questions please consider the [users group](https://groups.google.com/forum/#!forum/scikit-multiflow-users).

smastelini:
Hello @scikit-multiflow/scikit-multiflow-repo-admins,

I created this PR to fix #71. I removed the multiple `calc_byte_size_including_subtree` routines and the overwritten `__sizeof__` methods, changing them to single `calculate_object_size` calls at the root nodes. Besides, I also changed the `estimate_model_byte_size` method to be in conformity with the new way of calculating the object sizes. For the purpose of functionality, I've maintained the `measure_byte_size`, providing a way to externally measuring the tree size:

https://github.com/smastelini/scikit-multiflow/blob/1bb55df9889b3bb4463eb947401cabfd966a9417/src/skmultiflow/trees/hoeffding_tree.py#L859-L868

In addition, I also improved and corrected some packages' imports in `hoeffding_tree.py` file, which led me to perform some changes in other files too. I hope that those changes are not too confusing.

Thanks in advance for evaluating this PR.

Best regards!
Saulo Martiello Mastelini

Checklist

- [x] Travis CI build passes.


jacobmontiel:
Thanks @smastelini !

Just to confirm, the new approach is to use `calculate_object_size` which recursively calculates the size of objects. In the case of `HAT` it means that ADWIN is also considered.

I didn't understand the change  in `MTR-HT` regarding `ActiveLearningNodeForRegression`?

smastelini:
Hi @jacobmontiel,

Sorry for the confusion. You are right about the approach. Since the ADWIN object is a HAT's property, it is also considered when calculate the tree size.

The `ActiveLearningNodeForRegression` was added in the `MTR-HT` just for instantiating the right numerical attribute observer: the original inherited version from `HT` did not use the multi-target observer.

Sorry again for the additional features sent along with the original purposes addressed by this PR.

jacobmontiel:
Thanks @smastelini , it is clear now.

I will do the merge and add a note regarding the numerical attribute observer.

Thanks again for the support.

smastelini:
Thanks @jacobmontiel for the support and fast responses, as always!

PR #89, opened 2019-02-22
Add LearnNSE algorithm

caoquan95:
Fixes # .

Add LearnNSE algorithm
Compare to MOA: 
  SEAGenerator with window size of 1000:
    MOA mean acc: 83.38
    multiflow learnNSE: 95.58
  RandomTreeGenerator with window size of 1000:
    MOA mean acc: 68.12
    multiflow learnNSE: 63.17

The only difference is I use GaussianNaiveBayes and I am not sure which Naive Bayes MOA use

Checklist

- [x] Travis CI build passes.


jacobmontiel:
Hello @caoquan95 

`GaussianNaiveBayes` is slightly different from the incremental version in MOA. This could be the reason for the difference between the tests.

You could try using a different base estimator, for example, Hoeffding Trees.

jacobmontiel:
@caoquan95 

Notice that, to support both incremental and batch methods you need something like this:

``` python
try:
    classifier.partial_fit(self.X_batch, self.y_batch, sample_weight=self.instanceWeights)
except NotImplementedError:
    classifier.fit(self.X_batch, self.y_batch, sample_weight=self.instanceWeights)
```

caoquan95:
> Hello @caoquan95
> 
> `GaussianNaiveBayes` is slightly different from the incremental version in MOA. This could be the reason for the difference between the tests.
> 
> You could try using a different base estimator, for example, Hoeffding Trees.

I test with Hoeffding Tree, I got only 0.3 accuracy while MOA had 0.8. I still cannot find out what is the error. I implement base on the LearnNSE MOA

jacobmontiel:
@caoquan95 

I did a quick check and didn't find an error but I will check in more detail during the weekend. (btw nice usage of np instead of `for` loops)

Can you share the experiment setup you are using? There are different things to consider such as the chances that the data is different (specially if you are using a generator), model defaults in `MOA` vs `scikit-multiflow`, etc.

caoquan95:
@jacobmontiel 
My experiment is in the test_learn_nse.py file.

caoquan95:
@jacobmontiel 
Dear Sir,
I found the problem.
The weight parameter of the HoeffdingTree does not work. If I have the following code.

classifier = HoeffdingTree()

m = 250
# Pre training the classifier
X, y = stream.next_sample(m)
weight = np.ones(m) / m
classifier.partial_fit(X, y, weight=weight.tolist(), classes=stream.target_values)
print(classifier.predict(X))

It return

[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]

If I remove the weight, I got 0.9467 accuracy which is higher than MOA implementation

caoquan95:
I did some tests compare accuracy to MOA:
1. Test 1 SeaGenerator :
MOA LearnNSE Naive Bayes: 83.43
scikit multiflow LearnNSE Multinomial Naive Bayes: 75 
scikit multiflow LearnNSE Gaussian Naive Bayes: 91.44 

2 Test 2 SeaGenerator
MOA LearnNSE Hoeffding Tree: 85.6
scikit multiflow  LearnNSE Hoeffding Tree: 94.67

3. Test 3 RandomTreeGenerator :
MOA LearnNSE Naive Bayes: 68.16
scikit multiflow LearnNSE Multinomial Naive Bayes: 55.42 
scikit multiflow LearnNSE Gaussian Naive Bayes: 57.6 

4 Test 4 RandomTreeGenerator
MOA LearnNSE Hoeffding Tree: 67.76
scikit multiflow LearnNSE Hoeffding Tree: 66.46 

Here is the example code for experiment:
`stream = SEAGenerator()
stream.prepare_for_use()

estimator = GaussianNB()

classifier = LearnNSE(base_estimator=estimator)

# Keeping track of sample count and correct prediction count
sample_count = 0
corrects = 0

m = 1000
# Pre training the classifier
X, y = stream.next_sample(m)
classifier.partial_fit(X, y, classes=stream.target_values)

# print(classifier.ensemble_weights)
for i in range(10):
    X, y = stream.next_sample(m)
    pred = classifier.predict(X)
    classifier.partial_fit(X, y)

    if pred is not None:
        # print(pred)
        corrects += np.sum(y == pred)
    sample_count += m

acc = corrects / sample_count

print(acc)`

jacobmontiel:
Hello @caoquan95 

I have pushed into the master the Naive Bayes incremental learner (5bf139b5a4312e6bfbe6a196e68fb261ef674368) as it is implemented in MOA. Now you can use it as the default base estimator in Learn++.NSE.

I still get different results than in MOA. I will write in the review's comments about what I have found so far.

caoquan95:
> Hello @caoquan95
> 
> I have pushed into the master the Naive Bayes incremental learner ([5bf139b](https://github.com/scikit-multiflow/scikit-multiflow/commit/5bf139b5a4312e6bfbe6a196e68fb261ef674368)) as it is implemented in MOA. Now you can use it as the default base estimator in Learn++.NSE.
> 
> I still get different results than in MOA. I will write in the review's comments about what I have found so far.

@jacobmontiel 
I think the problem is that in MOA they support instances weight. But in scikit multiflow I cannot use the instances weight.

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+        self.y_batch = []
+        self.instance_weights = []
+        self.base_estimator = cp.deepcopy(base_estimator)
+        self.classes = None
+
+    def fit(self, X, y, classes=None, weight=None):
+        raise NotImplementedError
+
+    def __fit(self, estimator, X, y, sample_weight=None, classes=None):
+        try:
```
jacobmontiel: The `partial_fit` method is the default in `scikit_multiflow`, my suggestions are:

```python
    def __fit(self, estimator, X, y, classes=None, sample_weight=None):
        try:
            estimator.partial_fit(X, y, weight=sample_weight, classes=classes)
        except (NotImplementedError, TypeError):
            estimator.fit(X, y, weight=sample_weight, classes=classes)
```

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+                    et = np.sum(votes != self.y_batch) / mt
+
+                    # Update and normalize instance weights
+                    self.instance_weights = np.ones(mt) / mt
+                    self.instance_weights[votes == self.y_batch] = et / mt
+
+                    # normalize instance weights (distribution)
+                    self.instance_weights = self.instance_weights / np.sum(self.instance_weights)
+
+                    # Train base classifier with Dt
```
jacobmontiel: Weights are not used to train the ensemble member:

```python
# Train base classifier
self.__fit(classifier, self.X_batch, self.y_batch, sample_weight=None,
           classes=self.classes)
```

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+                    self.__fit(classifier, self.X_batch, self.y_batch, sample_weight=self.instance_weights,
+                               classes=self.classes)
+                    # classifier.fit(self.X_batch, self.y_batch, sample_weight=self.instanceWeights)
+
+                else:
+                    # First run! train the classifier on the instances with the same weight
+                    self.instance_weights = np.ones(mt) / mt
+
+                    # classifier.fit(self.X_batch, self.y_batch, sample_weight=self.instanceWeights)
```
jacobmontiel: Save comment as above, weights are not used to train the model `sample_weight=None`.

Also, please remove commented code.

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+
+                # Evaluate all existing classifiers on the new dataset
+                for k in range(1, t + 1):
+                    pred = self.ensemble[k - 1].predict(self.X_batch)
+                    ekt = np.sum(self.instance_weights[pred != self.y_batch])
+
+                    if k == t and ekt > 0.5:
+                        # Generate a new classifier
+                        classifier = cp.deepcopy(self.base_estimator)
+                        self.__fit(classifier, self.X_batch, self.y_batch, sample_weight=self.instance_weights,
```
jacobmontiel: The new classifier is not fitted at this point (in the MOA implementation). It seems strange so we need to confirm in the paper.

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+                    self.__fit(classifier, self.X_batch, self.y_batch, sample_weight=self.instance_weights,
+                               classes=self.classes)
+
+                self.ensemble.append(classifier)
+                self.bkts.append([])
+                self.wkts.append([])
+                self.ensemble_weights = []
+
+                t = len(self.ensemble)
+                max_error = None
```
jacobmontiel: Suggestion:

```python
max_error = -np.inf
error_index = -sys.maxsize - 1
```

Although `error_index` could be as simple as `-1`

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+
+                if self.pruning == 'age' and t > self.ensemble_size:
+                    # Age-based
+                    self.ensemble.pop(0)
+                    self.ensemble_weights.pop(0)
+                    self.bkts.pop(0)
+                    self.wkts.pop(0)
+                elif self.pruning == 'error' and t > self.ensemble_size:
+                    # Error-based
+                    self.ensemble.pop(error_index - 1)
```
jacobmontiel: This can result in an exception if `error_index` has not been set.

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def __vote_proba(self, X, t, classes):
+        res = []
+        for m in range(len(X)):
+            votes = np.zeros((1, len(classes)))
+            for i in range(t):
+                if self.ensemble_weights[i] > 0:
+                    h = self.ensemble[i]
+                    y_predicts = h.predict_proba(X[m].reshape(1, -1))
+
```
jacobmontiel: Missing the normalization step:

```python
y_predicts /= np.linalg.norm(y_predicts, ord=1, axis=1, keepdims=True)
```

src/skmultiflow/meta/learn_nse.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns
+        -------
+        numpy.ndarray
+            An array of shape (n_samples, n_features), in which each outer entry is
+            associated with the X entry of the same index. And where the list in
+            index [i] contains len(self.target_values) elements, each of which represents
+            the probability that the i-th sample of X belongs to a certain label.
+
+        """
```
jacobmontiel: Protection for the case when `predict` is called before fitting the model:

```python
        if not self.ensemble:
            return np.zeros((len(X), 1))
```

jacobmontiel:
@caoquan95 
> I think the problem is that in MOA they support instances weight. But in scikit multiflow I cannot use the instances weight.

Some methods in `scikit-multiflow` support weights, like `HoeffdingTrees` and `NaiveBayes`. However the instance weights in Learn++.NSE are not used to train the members of the ensemble but to monitor the ensemble errors.

src/skmultiflow/meta/learn_nse.py
caoquan95: If I don't fit it, it throws the error the classifier is not fitted when I run to the next iteration.

caoquan95: screenshot from 2019-02-25 13-00-47
Here is from the paper. The new classifier needs to be trained

jacobmontiel: This makes sense to me. Now we have to consider that our implementation will provide different results once this part of the code is reached (which in my experiments does).

We can use tests on a smaller scale to validate our implementation.

src/skmultiflow/meta/learn_nse.py
caoquan95: I don't know how to use weight in scikit multiflow. If I pass 1/m. It returns all 0. as I write when I use Hoeffding Tree

src/skmultiflow/meta/learn_nse.py
caoquan95: I think I've already passed in the instance_weights
self.__fit(classifier, self.X_batch, self.y_batch, sample_weight=self.instance_weights,

src/skmultiflow/meta/learn_nse.py
caoquan95: the error_index will always be set after we run the code
 ```
if ekt > max_error:
    max_error = ekt
    error_index = k
```

File: src/skmultiflow/meta/learn_nse.py (hunk 2 above)
Comment (caoquan95) [reply src/skmultiflow/meta/learn_nse.py]: screenshot from 2019-02-25 13-58-52
They use the weights to train the classifier

**Comment by caoquan95** [comment]
@jacobmontiel 
I will make some experiments with the Naive Bayes. 
Could you show me how to use the weight in scikit multiflow?

File: src/skmultiflow/meta/learn_nse.py (hunk 2 above)
Comment (jacobmontiel) [reply src/skmultiflow/meta/learn_nse.py]: The pseudo code is rather vague, from the text:
> Unlike most other ensemble algorithms, the instance error weights (i.e., the penalty distribution) in Learn++.NSE are not used for data (re)sampling or instance selection, but rather to weigh and assign error (later in Step 4). In fact, since the environment may change at any time, and that data are received in (possibly small) batches, all training data in are Dt
used for training, returning the hypothesis ht in Step 3.

File: src/skmultiflow/meta/learn_nse.py (hunk 1 above)
Comment (jacobmontiel) [reply src/skmultiflow/meta/learn_nse.py]: I reviewed the implementation of HoeffdingTrees and NaiveBayes and found no errors. What you describe is related to the fact that since we are normalizing the `weights` it means that the contribution of each sample is very small and the model is updated in small increments.

From your example above you can test this by changing:
```python
weight = np.ones(m) / (m//10)
```

Notice that there was indeed a bug in HoeffdingTree when passing more than 1 sample, but the fix is in place in the master. Please bring those changes to your branch.

jacobmontiel:
Here is the modification to your example using weights:

```python
from skmultiflow.data import AGRAWALGenerator
stream = AGRAWALGenerator()
stream.prepare_for_use()

classifier = NaiveBayes()
stream.restart()
m = 250
# Pre training the classifier
X, y = stream.next_sample(m)
weight = np.ones(m) / (m//10)
classifier.partial_fit(X, y, weight=weight)
X_test, y_test = stream.next_sample(20)
print('Predictions: ', classifier.predict(X_test))
print('Probabilities: ', classifier.predict_proba(X_test))
print('Observed class distribution: ', classifier._observed_class_distribution)
```

Notice that I use a different normalization factor in order to expose the model to "more" information.

caoquan95:
@jacobmontiel 
I use your Naive Bayes and compare with MOA.
```
stream = SEAGenerator(random_state=2212)
stream.prepare_for_use()

estimator = NaiveBayes()

classifier = LearnNSE(base_estimator=estimator)

# Keeping track of sample count and correct prediction count
sample_count = 0
corrects = 0

m = 1000
# Pre training the classifier
X, y = stream.next_sample(m)
classifier.partial_fit(X, y, classes=stream.target_values)

# print(classifier.ensemble_weights)
for i in range(10):
    X, y = stream.next_sample(m)
    pred = classifier.predict(X)
    classifier.partial_fit(X, y)

    if pred is not None:
        # print(pred)
        corrects += np.sum(y == pred)
    sample_count += m

acc = corrects/sample_count
print(acc)

```

I got 92.94 accuracy compare to 83.43 in MOA

caoquan95:
@jacobmontiel 
If I don't use the instance weights in the partial_fit and fit of the estimator. 
I have a better and closer accuracy to MOA

jacobmontiel:
I have been testing your implementation vs the one in MOA and so far I found two issues (in MOA):

- When a member of the ensemble is replaced but  it is not trained (as discussed previously)
- There is a bug in the creation of new instances of the base estimator (out of the scope of this project)

Taking these two issues into consideration I think that it is unreliable to use the MOA implementation as the reference. Let's continue with tests comparing Learn++.NSE (using NB as the base estimator) vs single NB. Once we are certain we could move to something like DT or HoeffdingTrees.

caoquan95:
> I have been testing your implementation vs the one in MOA and so far I found two issues (in MOA):
> 
> * When a member of the ensemble is replaced but  it is not trained (as discussed previously)
> * There is a bug in the creation of new instances of the base estimator (out of the scope of this project)
> 
> Taking these two issues into consideration I think that it is unreliable to use the MOA implementation as the reference. Let's continue with tests comparing Learn++.NSE (using NB as the base estimator) vs single NB. Once we are certain we could move to something like DT or HoeffdingTrees.

@jacobmontiel 
With SeaGenerator, I got 
Learn NSE 0.9467
Naive Bayes 0.9451

With RandomTreeGenerator, I got
Learn NSE 0.6069
Naive Bayes 0.6114

caoquan95:
screenshot from 2019-02-26 16-55-23

You are correct that they do not use the instance weights for training. The 2 notations D and D bold are different

jacobmontiel:
Thanks for the last changes @caoquan95 

Regarding the testing, the testing, it is not unexpected because:
1. The sample size is very small (in the context of data streams) so it is not clear how the classifiers perform over time
2. The data has no concept drift, which is the focus of Learn++.NSE

I ran two tests with your code:

## Electricity
~45K samples
Learn++.NSE with "age" based pruning
learn nse-elec

## Agrawal + gradual drift
1M samples
Learn++.NSE with "age" based pruning
learn nse-age-agr_g-1m

## Agrawal + gradual drift
1M samples
Learn++.NSE with "age" based pruning
learn nse-error-agr_g-1m

___

This is what we were expecting to see. Learn++.NSE reacts faster to concept drift.

I think we are ready to close this PR, let me know what do you think.

caoquan95:
@jacobmontiel 
Thank you very much. I agree to close this PR.
I spent the whole yesterday to see if I missed something from the paper and I found nothing.

jacobmontiel:
Thanks, @caoquan95 for the hard work up to this point.

PR #90, opened 2019-02-25
Add Hoeffding Anytime Tree & Docker. Fix #68 & #91. 

### LINKED ISSUE #68: test_adaptive_random_forests.py fails on Python 3.6+
### Expected behaviour
The test should pass on Python 3.4.x, 3.5.x and 3.6.x

### Actual behaviour
Test files on Python 3.6.x while passing on 3.4.x and 3.5.x (locally and in Travis)

### Steps to reproduce the behaviour
```
pytest tests/meta/test_adaptive_random_forests.py -v
```

Additional information:
This issue raised when working on #66. This issue is intended to further investigate the the root cause of the problem.

The assert function was **temporary** disabled for Python 3.6.x in 838cb91565c0bc029919796b7cdc99e31602b9a7
This issue will re-enable the test for all Python versions supported.

garawalid:
- Implementation of Hoeffding Anytime Tree ([Extremely Fast Decision Tree](https://arxiv.org/pdf/1802.08780.pdf)).
- Create Docker organization for scikit-multiflow: [skmultiflow/scikit-multiflow](https://hub.docker.com/r/skmultiflow/scikit-multiflow).
- Dockerfiles for multiple tags: 

Tag | details |
--- | --- | 
latest | python image follows pypi release
jupyter  | jupyter image follows pypi release
devel |  python image tracks Github repository
devel-jupyter | jupyter image that tracks Github repository

What I mean by tracking repository is, whenever a new commit happens in `scikit-multiflow` repository, Docker Hub will rebuild the images. 

N.B: The first two images need to be updated manually, so I volunteer to maintain them. Also, it's possible to build and push images with the makefile (`make build-jupyter` and `make push-jupyter`)

- Add documentation for scikit-multiflow Docker image with hello world example. Also, [skmultiflow/scikit-multiflow](https://hub.docker.com/r/skmultiflow/scikit-multiflow) contains a tutorial how to use scikit-multiflow container.
- Add some useful methods in `hoeffding_tree.py`.

Discussion about information gain in [MOA](https://github.com/Waikato/moa) and scikit-multiflow : 
------
Both MOA and scikit-multiflow compute the information gain of don't split (null attribute or for short X_null) the same way, they retuœrn `- infinity'.

Code from: `info_gain_split_criterion.py`
```python
    def get_merit_of_split(self, pre_split_dist, post_split_dist):
        if self.num_subsets_greater_than_frac(post_split_dist, self.min_branch_frac_option) < 2:
            return -np.inf

        return self.compute_entropy(pre_split_dist) - self.compute_entropy(post_split_dist)
```
Since Hoeffding Tree ([Mining High-Speed Data Streams](https://homes.cs.washington.edu/~pedrod/papers/kdd00.pdf)) always takes the two best attribute in terms of information gain (G) and then compare them to Hoeffding bound (ε) : 
```
If G(X_a) - G(X_b) >  ε and X_a != X_null then split. 
```
But in Hoeffding Anytime Tree (HATT), we attempt to split a node based on this condition : 
```
If G(X_a) - G(X_null) >  ε and X_a != X_null then split.
```
In Hoeffding Tree, we are not dealing with `X_null`, that's why we will never get `G(X_null) = -inf`, meanwhile, all the attempt to split method in HATT will be affected by `G(X_null) = -inf`. So the previous condition will always hold : `+inf >  ε`.

Here is an example when we use `G(X_null) = -inf`
 
**M0 : HoeffdingTree**  
**M1 : HoeffdingAnytimeTree**
```
Dataset : RandomTreeGenerator
n_classes=2  
n_categorical_features = 8  
n_num_features = 1  

Pre-training on 10 sample(s).
Evaluating...
 #################### [100%] [159.31s]
Processed samples: 15000
Mean performance:
M0 - Accuracy     : 0.5502
M0 - Kappa        : 0.0886
M1 - Accuracy     : 0.5789
M1 - Kappa        : 0.1450
```

exp1_acc

exp1_nb_an

As you can see, the number of active nodes (number of leaves) grows quickly and can reach around 240 leaves after approximately 3200 samples processed. And the accuracy quite higher than HT. But there are some other cases where HT performs better. 

In this example we will generate numerical features, here is the result: 
```
Dataset : RandomTreeGenerator
n_classes=2
n_categorical_features = 0
n_num_features = 12

Prequential Evaluation
Evaluating 1 target(s).
Pre-training on 10 sample(s).
Evaluating...
 #################### [100%] [1062.53s]
Processed samples: 15000
Mean performance:
M0 - Accuracy     : 0.6942
M0 - Kappa        : 0.2651
M1 - Accuracy     : 0.5825
M1 - Kappa        : 0.1343
```
exp3_acc
exp3_nb_an

Fix proposed for `G(X_null) = - inf`
------
Theoretically, `G(X_null)` should be equal to zero. I think the goal of `G(X_null) = - inf` in MOA and scikit-multiflow, is to make sure that the null attribute is the last element when we sort attributes based on information gain. Especially, when we have multiple attributes that have the information gain equals to zero.
To fix this, I forced `G(X_null)` to be zero instead `-inf`
```python
# Force x_null merit to get 0 instead of -infinity
if x_null.merit == -np.inf:
    x_null.merit = 0.0
```
If we take the previous example and see the changes, here is how it looks like : 

```
Dataset : RandomTreeGenerator
n_classes=2
n_categorical_features = 0
n_num_features = 12

N.B: Information gain of null split get zero instead of -inf

Pre-training on 10 sample(s).
Evaluating...
 #################### [100%] [438.89s]
Processed samples: 15000
Mean performance:
M0 - Accuracy     : 0.6942
M0 - Kappa        : 0.2651
M1 - Accuracy     : 0.6701
M1 - Kappa        : 0.2325
```

exp4_acc
exp4_nb_an

Also, I did other tests on different datasets : 

**Summary:** 

Dataset | HT | HAT | HATT (0) | HATT (-inf)
--- | --- | --- | --- | --- 
Waveform | 0.8015 | 0.7932 | 0.8152 | 0.7936
Electricity | 0.8276 | 0.8822 | 0.8372 | 0.8524
SeaGenerator | 0.8025 | 0.7938 | 0.8091 | 0.7988

HAT: Hoeffding Adaptive Tree
HT: Hoeffding Tree
HATT (0): Hoeffding Anytime Tree where `G(X_null) = 0`
HATT (-inf): Hoeffding Anytime Treewhere `G(X_null) = -inf`

I think it's better to override the value of information gain of "don't split attribute" and we use `G(X_null) = 0`. Also, we have to take into account that when `G(X_null) = -inf`, the tree goes big and can reach 300 leaves in a short time, so this will increase both memory and time processing. 
@jacobmontiel what do you think?

Checklist

- [x] Travis CI build passes.


jacobmontiel:
@garawalid 

Thanks for the detailed description!

Here are my comments:

### Docker
*Disclaimer:* I am not familiar with the details of docker.
- Is it necessary to keep the Docker folder as part of the main project? Another way to phrase this, can we place it somewhere else under the scikit-multiflow organization on GitHub? For example  https://github.com/scikit-multiflow/docker
- If it is possible to relocate it, does it make sense?
- Do not copy `elec.csv`. Better rely only on stream generators for examples. We will remove the datasets from the main project in the future.

### HATT
- I agree that `G(X_null) = 0` seems to be the best option, and if I recall correctly this is what the paper assumes.
- The test is [failing on travis](https://travis-ci.org/scikit-multiflow/scikit-multiflow/builds/498272653) for python 3.4/3.5

### Misc
- Could you elaborate on:
>     * Add some useful methods in `hoeffding_tree.py`
- I saw the implementation of the methods, should they be rather set as properties?
- Also related to this, the docstring on these methods is not well formed.

garawalid:
@jacobmontiel 
Thanks for your comments ! 

Docker
--------
+ Usually they put docker files in the main project, for example [TensorFlow](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/dockerfiles) & [PyTorch](https://github.com/pytorch/pytorch/tree/master/docker). Also, we need the dockerfile to be in the same repository, so whenever a new commit happens, DockerHub will trigger auto build for `scikit-multiflow:devel` and `scikit-multiflow:devel-jupyter`
+ In the scikit-multiflow docker image, I add some hello world example, both `ht_from_file.py` and `QuickStart.ipynb` use `elec.csv`. The idea behind it is to give an example of how to use csv as a source of data. If we can download  `elec.csv`, then it's possible to delete it. Just, I'll download it inside the docker image when we build it.

HATT
--------
+ Exactly, the paper assume that `G(X_null) = 0`. 
+ I fixed the test for both python 3.4 & 3.5. Processing 15000 samples gives a different output that depends on python version. 

Misc
--------
+ I tried to make these methods properties, the code failed to call them from `hoeffding_anytime_tree.py`. I think it's better that we keep them as methods.
+ I fixed the docstring.

garawalid:
@jacobmontiel 

Docker
--------
I made a python script `data.py` that downloads the [electricity dataset](https://www.openml.org/d/151), then it drops the two first columns and performs label encoding for the target column. This script will be executed within the docker image while building it. This is how we avoid keeping `elec.csv` in the repository.

jacobmontiel:
Thanks @garawalid 

> I fixed the test for both python 3.4 & 3.5. Processing 15000 samples gives a different output that depends on python version.

This is very strange, do you have any idea about why they are different? Is this an issue from the generator (different data) or the classifier (different model)?

> I made a python script `data.py` that downloads the electricity dataset

This is what I had in mind and likely what we will be doing in the future.

garawalid:
@jacobmontiel 

> This is very strange, do you have any idea about why they are different? Is this an issue from the generator (different data) or the classifier (different model)?

You are right about the strange behavior, I verified the `RandomTreeGenerator`, it returns the same stream data for both python3.5 &3.6.  Also, the classifier has the same model in python 3.4, 3.5 & 3.6, but there is a difference in the probability and predictions. 

Also, I get **exactly** the same probability for python3.4 & 3.5. These results are the same for `Hoeffding Tree` and `HATT`.

I suggest that we discuss it in this issue #91

jacobmontiel:
Agree. Thanks for checking this and opening the issue to track it.

garawalid:
Fix issue #68 and #91. 
HATT test works with Python 3.7.2.

jacobmontiel:
Thanks for the fixes @garawalid 

Can you activate the `assert`s in the ARF tests to validate the fix for #68 ?

garawalid:
@jacobmontiel 
Do you mean that I need to add more `assert` in `test_adaptive_random_forests`?
Because the 3 `assert`s are activated in commit 1e7901d.

jacobmontiel:
Yes, those are the ones, sorry I missed the change to those tests.

jacobmontiel:
Add Hoeffding Anytime Tree
Add Docker support
Fix #68 
Fix #91
Minor changes and updates

PR #94, opened 2019-03-16
Update LearnNSE meta classifier implementation

### LINKED ISSUE #93: LearnNSE implementation does not work on classifiers without partial_fit
LearnNSE implementation does not work on classifiers without partial_fit.

It is caused by possible bug in the code, since `.fit(...)` method on classifier will never be executed when using classifiers from current version of scikit-learn api. 

https://github.com/scikit-multiflow/scikit-multiflow/blob/ddf104437132f3b75f3fa6d30195e09d7bcb3231/src/skmultiflow/meta/learn_nse.py#L69-L73

`estimator.fit(X, y, classes=classes) ` will always raise *TypeError* because _classes_ property is not known (it is in `.partial_fit`)

bgulowaty:
this pull request fixes #93

### DIFF
--- a/src/skmultiflow/meta/learn_nse.py
+++ b/src/skmultiflow/meta/learn_nse.py
@@ -1,7 +1,8 @@
 import copy as cp
 import numpy as np
-from skmultiflow.core.base import StreamModel
 from sklearn.tree import DecisionTreeClassifier
+
+from skmultiflow.core.base import StreamModel
 
 
 class LearnNSE(StreamModel):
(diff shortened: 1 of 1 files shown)


src/skmultiflow/meta/learn_nse.py
```diff
@@ -1,8 +1,9 @@
 import copy as cp
 import numpy as np
-from skmultiflow.core.base import StreamModel
 from sklearn.tree import DecisionTreeClassifier
 
+from skmultiflow.core.base import StreamModel
```
jacobmontiel: This change is unnecessary, can you keep the original imports? Just for clarity.

bgulowaty: It was automatically reformated by my editor.
In PEP-8 you can read https://www.python.org/dev/peps/pep-0008/#imports
```

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

You should put a blank line between each group of imports.
```

I think its good change

jacobmontiel: I see your point and agree. I will merge the PR.

Thanks for the help @bgulowaty

PR #101, opened 2019-04-05
Feature Additive Expert Ensemble

AndreFCruz:
I'm trying to implement the _Additive Expert Ensemble_ algorithm, but keep facing an error:
- the NaiveBayes estimator (base_estimator instance) returns probabilities whose sum is 0;
- therefore, line 172 of naive_bayes.py leads to a division by zero "y_proba[int(key)] = value / sum_values" ;

Screen Shot 2019-04-05 at 16 17 25

Any idea why this could be happening? (given that this file was not changed and I only use the classifier's predict method)

- Implementation of Additive Expert Ensemble algorithm [1]
[1] Kolter and Maloof. Using additive expert ensembles to cope with Concept drift. Proc. 22 International Conference on Machine Learning, 2005.


jacobmontiel:
Hello @AndreFCruz 

I suspect that `do_naive_bayes_prediction` is returning {0:0.0} which can happen if the observed distribution is empty, which shouldn't be in this case given that you are training on 100 samples before the error. Could you share an MWE?

Additionally to this, the code is not properly protected against division by zero. I will submit a fix after figuring the issue you are reporting (i.e. why is the sum zero?)

AndreFCruz:
Thanks for the feedback, I think I found a couple bugs on my code.
I made a couple changes (_e.g._ guarding against predictions when experts aren't yet trained), and now I only see the error once in the beginning (which is still odd given that it's after pretraining):
Screen Shot 2019-04-05 at 23 32 06


Here is a MWE:
```python
from skmultiflow.bayes import NaiveBayes
from skmultiflow.evaluation import EvaluatePrequential
from skmultiflow.data import WaveformGenerator

stream = WaveformGenerator()
stream.prepare_for_use()
evaluator = EvaluatePrequential(show_plot=True,
                                pretrain_size=200,
                                max_samples=5000)

from skmultiflow.meta import AdditiveExpertEnsemble
ensemble = AdditiveExpertEnsemble(max_estimators=5, base_estimator=NaiveBayes())

evaluator.evaluate(stream=stream, model=ensemble)
```

xfontes42:
I think the problem lies somewhere in this portion of the code from the file `utils.py`. Following up the `sum_values` variable in the `naive_bayes.py` file.
```python
if attribute_observers:
    for att_idx in range(len(X)):
        if att_idx in attribute_observers:
            obs = attribute_observers[att_idx]
            votes[class_index] *= obs.probability_of_attribute_value_given_class(X[att_idx], class_index)
```

jacobmontiel:
Thanks for the info @AndreFCruz and @xfontes42 

I looked in detail and found that:

1) The zero value is indeed generated by `do_naive_bayes_prediction`, but not by an empty observed distribution. The actual reason is the zero variance when the attribute observers have not seen enough data. However, this is the expected behavior in `do_naive_bayes_prediction`.
2) The issue is not triggered by the model trained in the first 200 samples but by a new `expert` model introduced in the ensemble. Since this new model has not seen enough data then it returns zero votes as explained above.
3) Taken this into consideration, the bahiour of `NaiveBayes` is correct and we just need to address properly the division by zero. I will commit a fix in the `predict_proba` of `NaiveBayes` to properly handle this when normalizing the votes.

*Note:* During this analysis I noticed that the `WaveformGenerator` yields class labels as `float` when they should be `int`. This has no impact on the reported issue but I will address it as well in a fix commit.

Once again thanks for the report.

jacobmontiel:
The fix is available in the master branch (f7bad21abd5a7fc15f150c180eb70916a81013bf)

AndreFCruz:
Thanks,
I still have a couple thing I'd like to improve on this implementation of AdditiveExpertEnsemble, so I'll close this PR and open non-draft PR soon.

PR #102, opened 2019-04-08
Feature/dynamic weighted majority

AndreFCruz:
- Implementation of the Dynamic Weighted Majority ensemble algorithm [1]

[1] Kolter and Maloof. Dynamic weighted majority: An ensemble method for drifting concepts. The Journal of Machine Learning Research, 8:2755-2790, December 2007. ISSN 1532-4435.

Loosely based on the MOA implementation of the same algorithm: https://github.com/Waikato/moa/blob/master/moa/src/main/java/moa/classifiers/meta/DynamicWeightedMajority.java

_Note:_ this is a different algorithm from #101 

Checklist

- [x] Travis CI build passes.

### DIFF
--- a/src/skmultiflow/meta/dynamic_weighted_majority.py
+++ b/src/skmultiflow/meta/dynamic_weighted_majority.py
@@ -122,27 +122,13 @@ class DynamicWeightedMajority(StreamModel):
         numpy.ndarray
             A numpy.ndarray with the label prediction for all the samples in X.
         """
-        return np.array([np.argmax(self.predict_proba(X))])
+        preds = np.array([np.array(exp.estimator.predict(X)) * exp.weight
+                          for exp in self.experts])
+        sum_weights = sum(exp.weight for exp in self.experts)
+        return np.sum(preds / sum_weights, axis=0, dtype=int)
 
     def predict_proba(self, X):
-        """ predict_proba
-
-        Predicts the probability of each sample belonging to each one of the
-        known classes.
-
-        Parameters
-        ----------
-        X: Numpy.ndarray of shape (n_samples, n_features)
-            A matrix of the samples we want to predict.
-
-        Returns
-        -------
-        numpy.ndarray
-            An array of shape (n_samples, n_features), in which each outer
-            entry is associated with the X entry of the same index.
-        """
-        return self._aggregate_expert_predictions(
-                    self.get_expert_predictions(X))
+        raise NotImplementedError
 
     def fit_single_sample(self, X, y, classes=None, weight=None):
         """
@@ -206,10 +192,10 @@ class DynamicWeightedMajority(StreamModel):
         for exp in self.experts:
             exp.estimator.partial_fit(X, y, classes, weight)
 
-    def get_expert_predictions(self, X, classes=None, weight=None):
+    def get_expert_predictions(self, X):
         """
         Returns predictions of each class for each expert.
-        In shape: (n_experts,)
+        In shape: (n_experts, n_samples)
         """
         return [exp.estimator.predict(X) for exp in self.experts]
 
@@ -230,17 +216,6 @@ class DynamicWeightedMajority(StreamModel):
         scale_factor = 1 / max_weight
         for exp in self.experts:
             exp.weight *= scale_factor
-
-    def _aggregate_expert_predictions(self, predictions):
-        """
-        Aggregate predictions of all experts according to their weights.
-        Returns array of shape: (n_classes,)
-        """
-        aggregate_preds = np.zeros((np.max(predictions) + 1,))
-        for pred, w in zip(predictions, (exp.weight for exp in self.experts)):
-            aggregate_preds[pred] += w
-
-        return aggregate_preds / sum(exp.weight for exp in self.experts)
 
     def _remove_experts(self):
         """
--- a/src/skmultiflow/meta/dynamic_weighted_majority.py
+++ b/src/skmultiflow/meta/dynamic_weighted_majority.py
@@ -71,25 +71,25 @@ class DynamicWeightedMajority(StreamModel):
         self.reset()
 
     def fit(self, X, y, classes=None, weight=None):
-        raise NotImplementedError        
+        raise NotImplementedError
 
     def partial_fit(self, X, y, classes=None, weight=None):
         """
         Partially fits the model on the supplied X and y matrices.
 
-        Since it's an ensemble learner, if X and y matrix of more than one 
-        sample are passed, the algorithm will partial fit the model one sample 
+        Since it's an ensemble learner, if X and y matrix of more than one
+        sample are passed, the algorithm will partial fit the model one sample
         at a time.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


src/skmultiflow/meta/dynamic_weighted_majority.py
```diff
@@ -0,0 +1,271 @@
...
+            The estimator to wrap.
+        weight: float
+            The estimator's weight.
+        """
+        def __init__(self, estimator, weight):
+            self.estimator = estimator
+            self.weight = weight
+
+        def __lt__(self, other):
+            self.weight < other.weight
```
jacobmontiel: Missing `return`?

```suggestion
            return self.weight < other.weight
```

src/skmultiflow/meta/dynamic_weighted_majority.py
```diff
@@ -0,0 +1,271 @@
...
+            
+        classes: list 
+            List of all existing classes. This is an optional parameter.
+
+        weight: None
+            Instance weight. This is ignored by the ensemble and is only
+            for compliance with the general skmultiflow interface.
+        """
+        self.epochs += 1
+        self.num_classes = max(
```
jacobmontiel: `self.num_classes` is not defined in the `__init__` method

src/skmultiflow/meta/dynamic_weighted_majority.py
```diff
@@ -0,0 +1,271 @@
...
+            if np.any(yHat != y):
+                if len(self.experts) == self.max_experts:
+                    self.experts.pop(weakest_expert_index)
+                self.experts.append(self._construct_new_expert())
+
+        # Train individual experts
+        for exp in self.experts:
+            exp.estimator.partial_fit(X, y, classes, weight)
+
+    def get_expert_predictions(self, X, classes=None, weight=None):
```
jacobmontiel: Parameters `classes` and `weight` are not used in this method

src/skmultiflow/meta/dynamic_weighted_majority.py
```diff
@@ -0,0 +1,271 @@
...
+        self.num_classes = max(
+            len(classes) if classes is not None else 0,
+            (int(np.max(y)) + 1), self.num_classes)
+        predictions = np.zeros((self.num_classes,))
+        max_weight = 0
+        weakest_expert_weight = 1
+        weakest_expert_index = None
+        
+        for i, exp in enumerate(self.experts):
+            yHat = exp.estimator.predict(X)
```
jacobmontiel: The variable name should be lowercase `y_hat`

src/skmultiflow/meta/dynamic_weighted_majority.py
```diff
@@ -0,0 +1,271 @@
...
+        """
+        Reset this ensemble learner.
+        """
+        self.epochs = 0
+        self.num_classes = 2    # Minimum of 2 classes
+        self.experts = [
+            self._construct_new_expert()
+        ]
+
+    def _scale_weights(self, maxWeight):
```
jacobmontiel: The variables names should be lowercase: `max_weight`, `scale_factor`

src/skmultiflow/meta/dynamic_weighted_majority.py
```diff
@@ -0,0 +1,271 @@
...
+        Returns predictions of each class for each expert.
+        In shape: (n_experts,)
+        """
+        return [exp.estimator.predict(X) for exp in self.experts]
+
+    def reset(self):
+        """
+        Reset this ensemble learner.
+        """
+        self.epochs = 0
```
jacobmontiel: Some attributes are defined outside the `__init__` method:
`self.epochs`, `self.num_classes`, `self.experts`

tests/meta/test_dynamic_weighted_majority.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            learner.partial_fit(X, y)
+        cnt += 1
+    performance = correct_predictions / len(predictions)
+    expected_predictions = [1, 0, 1, 1, 0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 1, 1, 1,
+                            0, 1, 0, 0, 1, 0, 0, 1, 1, 1, 1, 0, 1, 0, 1, 1, 1,
+                            0, 0, 1, 1, 1, 0, 0, 1, 1, 0, 1, 0, 0, 1, 1]
+    expected_correct_predictions = 44
+    expected_performance = 0.8979591836734694
+
```
jacobmontiel: Given that the method implements `predict_proba` please consider adding an explicit assertion for expected predicted probabilities.

Although `predict_proba` is called by `predict`, there could be cases where changes in `predict_proba` could be missed without an explicit test.

See for reference: https://github.com/scikit-multiflow/scikit-multiflow/blob/c4dbbb70d4ed839d95a18ca799f073ac9ff9ba49/tests/bayes/test_naive_bayes.py#L44-L46

src/skmultiflow/meta/dynamic_weighted_majority.py
AndreFCruz: should I repeat the body of ```reset()``` in ```__init__``` ? (instead of calling reset in __init__)

jacobmontiel: In my opinion, both are correct. Personally, I prefer calling `reset` from `__init__`.

tests/meta/test_dynamic_weighted_majority.py
AndreFCruz: The ```predict_proba``` was a mistake, it was just to be used internally since the paper doesn't actually define what the class probability distributions would be.

I've addressed this in the most recent commit.

src/skmultiflow/meta/dynamic_weighted_majority.py
AndreFCruz: it's defined in the ```reset()``` method, which is called in ```__init__```

AndreFCruz:
Thanks for the feedback

A couple notes:
* some variables are not defined in ```__init__```, but are defined in the ```reset``` method which is called in ```__init__```
* ```predict_proba``` is no longer implemented
* removed the ```__lt__``` method from the inner class, this wasn't being used
* removed unused optional args
* changed variable naming to snake_case

jacobmontiel:
Thanks for the changes @AndreFCruz 

Regarding the instance attributes. This is just for readability, so people unfamiliar with the code can "see" all attributes in the `__init__`. For example, in this case:

``` python
    def __init__(self, n_estimators=5, base_estimator=NaiveBayes(),
                 period=50, beta=0.5, theta=0.01):
        """
        Creates a new instance of DynamicWeightedMajority.
        """
        super().__init__()

        self.max_experts = n_estimators
        self.base_estimator = base_estimator

        self.beta = beta
        self.theta = theta
        self.period = period

        # Following attributes are set later
        self.epoch = None
        self.num_classes = None
        self.experts = None

        self.reset()
```

Let me know what you think.

AndreFCruz:
Sure, I agree that improves readability.
Just committed the changes.

PR #110, opened 2019-05-04
Improved base class

### LINKED ISSUE #22: Automate get_info() method in base abstract object
### Expected behaviour
`get_info()` should be automatically generated.
See: https://docs.python.org/3/library/inspect.html

### Actual behaviour
Currently, `get_info()` is an abstract method that needs to be implemented manually across `scikit-multiflow`

### LINKED ISSUE #45: Models are not fully scikit-learn compatible
In order to follow scikit-learn's philosophy, wouldn't it make sense to derive your base model https://github.com/scikit-multiflow/scikit-multiflow/blob/c16892a053c6f033c4408ab9e418607b8b764919/src/skmultiflow/core/base.py#L5 

from the sklearn base class for estimators http://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html ?

Since `get_params` and `set_params` are not implemented in scikit-multiflow, useful sklearn utils like http://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html can not be applied to the scikit-multiflow models.

jacobmontiel:

* An improved base class that is compatible with `sklearn` `BaseEstimator`
* The new base class keeps the `reset` and `get_info` methods used across skmultiflow
* Specific mixin created: `ClassifierMixi`, `RegressorMixi`, `MetaEstimatorMixin` and `MultiOutputMixin`
* Some methods signatures were updated for consistency
* The `weight` attribute in `partial_fit` is now `sample_weight` for compatibility with `sklearn` methods
* Code documentation is updated
* Tests are updated
* Fixes #45 and fixes #22 

---

Checklist


src/skmultiflow/drift_detection/base_drift_detector.py
```diff
@@ -1,8 +1,8 @@
 from abc import ABCMeta, abstractmethod
-from skmultiflow.core.base_object import BaseObject
+from skmultiflow.core import BaseStreamEstimator
 
 
-class BaseDriftDetector(BaseObject, metaclass=ABCMeta):
+class BaseDriftDetector(BaseStreamEstimator, metaclass=ABCMeta):
```
remiadon: Is a `DriftDetector` really a `StreamEstimator` ? 
If I try to implement its `partial_fit` and `predict`methods, it gives something like:
```python
def partial_fit(self, X):
    for value in X:
        self.add_element(X)
    return self

def predict(self):
    return self.detected_change()   
```

with no input for predict I don't think it is an "estimator"

src/skmultiflow/data/base_stream.py
```diff
@@ -1,8 +1,8 @@
 from abc import ABCMeta, abstractmethod
-from skmultiflow.core.base_object import BaseObject
+from skmultiflow.core import BaseStreamEstimator
 
 
-class Stream(BaseObject, metaclass=ABCMeta):
+class Stream(BaseStreamEstimator, metaclass=ABCMeta):
```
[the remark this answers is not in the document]
jacobmontiel: `BaseObject` is a legacy class whose functionality is covered by the new `BaseStreamEstimator` which extends `BaseEstimator`

I agree that the naming is confusing though.

Current version:
`BaseObject` -> `StreamModel`

Proposed change
`BaseEstimator` -> `BaseStreamEstimator`

Some of the `StreamModel` functionality is now broken into task-specific mixins. For example, to define a single-target classifier:
Current version:
```python
class a_classifier(StreamModel)
```
Proposed change:
```python
class a_classifier(BaseStreamEstimator, ClassifierMixin)
```

src/skmultiflow/drift_detection/base_drift_detector.py
jacobmontiel: `fit`, `partial_fit, `predict` and `predict_proba` are now moved into task-specific mixins.

In this case, making the `BaseDriftDetector` class an extension of `BaseStreamEstimator` does not imply this extra functionality.

However, as previously mentioned, the name is confusing. We might rename `BaseStreamEstimator` to `BaseObject` for clarity.

jacobmontiel:
Thanks for looking into this

> I don't see the class definition for `BaseStreamEstimator` in the Files Changed, maybe I missed something ? 

Here it is:
https://github.com/scikit-multiflow/scikit-multiflow/blob/7e2d935a9387cfc6bf7fe1853e6f904e767c89b4/src/skmultiflow/core/base.py#L337-L364

> Second I think some changes in this PR are actually not related to the subject : "Improve base class"
> Those extra changes make the PR a bit big, and other PR more prone to present conflicts with this one
 
Changes are a side-effect from the new base class:
* `score` is now properly defined in each mixin so no need to define it in extended classes
* `fit` is now defined by `ClassifierMixin` and `RegressorMixin` by calling `partial_fit`. Previously, most methods where rising a `NotImplementedMethod`. On the other hand, if a better implementation is possible, then the `fit` method can be overriden.
* `get_info` is now based on `__repr__()` and `get_params` so no need to define it in extended classes
* Some changes were needed to address bugs where parameters were not fully visible to `get_params` which affects `get_info` and `clone`
* `fit` and `partial_fit` in the new mixins, parameter `weight` is renamed to `sample_weight` for compatibility with `sklearn`

I am trying to minimize the impact for users. However impact for developers (other PRs) is unavoidable.

Let me know if you have more questions/comments.

remiadon:
No more questions. Everything seems OK to me !

jacobmontiel:
This PR is ready. Code is complete and checks/tests pass.

PR #113, opened 2019-05-11
Streamz as input

remiadon:

* Creation of interfaces between `skmultiflow.data` and `streamz.Source` to make skmultiflow data generators as inputs to streamz.Stream(s)
* [TODO] add methods for triggering partial_fit/predict on skmultiflow estimators
---

Note that this PR also deprecate some existing classes, in order to encourage usage of streamz as the default way of handling streaming

Checklist

- [X] Tests are included for new functionality or updated accordingly.

This PR introduces new syntax 
```python
from streamz import Source
from skmultiflow.data import SEAGenerator
gen = SEAGenerator(random_state=1)
stream = Source.from_generator(gen, poll_interval=.5, batch_size=10) # pull 10 elements every 0.5 seconds
stream.sink(print) 
stream.start() # printing data from the SEAGenerator
```

# TODO : show syntax for partial_fit/predict


jacobmontiel:
Travis fails due to the latest version of sklearn. I will update a temporary workaround as part of the other open PR. I will upload the actual fix later today.

src/skmultiflow/core/stream.py
```diff
@@ -0,0 +1,28 @@
+from streamz import Stream
+
+@Stream.register_api()
+class predict(Stream):
```
jacobmontiel: This is a method reserved for learning methods.

src/skmultiflow/core/stream.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, upstream, model, **kwargs):
+        self.model = model
+        super().__init__(upstream, **kwargs)
+
+    def update(self, X, who=None):
+        return self._emit(self.model.predict(X))
+
+@Stream.register_api()
+class partial_fit(Stream):
```
jacobmontiel: This is a method reserved for learning methods.

src/skmultiflow/data/base_generator.py
```diff
@@ -2,26 +2,26 @@
 from skmultiflow.core.base_object import BaseObject
 
 
-class Stream(BaseObject, metaclass=ABCMeta):
-    """ The abstract class setting up the minimum requirements of a stream,
+class BaseGenerator(BaseObject, metaclass=ABCMeta):
```
jacobmontiel: This renaming could raise some confusion given that not all extended-classes are generators. Suggestions: `SMKStream`, `BaseStream`, `StreamMixin`,...

This is not limited to available methods, but also considering future sources such as kafka, tcp, http, etc.

src/skmultiflow/data/base_generator.py
```diff
@@ -2,26 +2,26 @@
...
         self.n_samples = 0
         self.n_targets = 0
         self.n_features = 0
         self.n_num_features = 0
         self.n_cat_features = 0
         self.n_classes = 0
-        self.cat_features_idx = []
-        self.current_sample_x = None
-        self.current_sample_y = None
+        self.cat_features_idx = []  # TODO : remove
```
jacobmontiel: Won't remove given that `cat_features_idx` is used by some learning methods that implement specific ways to process categorical features.

src/skmultiflow/data/data_generator.py
```diff
@@ -0,0 +1,74 @@
...
+import numpy as np
+from skmultiflow.data.base_generator import BaseGenerator
+
+# TODO : doc for all methods
+
+class DataGenerator(BaseGenerator):
+    # TODO : inherit from pandas.DataFrame should make things easier
+    """ DataGenerator
+
+    A generator constructed from the entries of a static dataset (numpy array or pandas
```
jacobmontiel: I like the idea of having a single method for both **file** and **raw data** sources. However, it would be nice to make clear in the example section how both functionalities remain. This is because a lot of users rely on these methods. Optionally we could keep `DataStream` and `FileStreams` as wrappers with a deprecation warning.

src/skmultiflow/data/stream.py
```diff
@@ -0,0 +1,30 @@
+import pandas as pd
+from tornado import gen
+from streamz import Stream, Source
+from skmultiflow.data.base_generator import BaseGenerator
+
+@Stream.register_api(staticmethod)
+class from_generator(Source):
```
jacobmontiel: Same comment about naming as in the `BaseGenerator` class.

src/skmultiflow/transform/missing_values_cleaner.py
```diff
@@ -1,197 +0,0 @@
-import numpy as np
-from scipy import stats
-from skmultiflow.transform.base_transform import StreamTransform
-from skmultiflow.utils.data_structures import FastBuffer
-from skmultiflow.utils.utils import get_dimensions
-
-
-class MissingValuesCleaner(StreamTransform):
```
jacobmontiel: Instead of removing, use a deprecation warning to let the users know that this will be removed and the alternative (sklearn.impute)

src/skmultiflow/transform/one_hot_to_categorical.py
```diff
@@ -1,23 +1,24 @@
 import numpy as np
-from skmultiflow.transform.base_transform import StreamTransform
 from skmultiflow.utils.utils import get_dimensions
 
+from sklearn.base import BaseEstimator, TransformerMixin
```
jacobmontiel: Nice to see that `TransformerMixin` works :+1:

src/skmultiflow/core/stream.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, upstream, model, **kwargs):
+        self.model = model
+        super().__init__(upstream, **kwargs)
+
+    def update(self, x, who=None):
+        X, y = X
+        self.model.partial_fit(X, y)
+        return self._emit(self.model)
```
martindurant: Here is what I was talking about. I guess it's OK to emit the whole model (in-process, anyway), and then downstream streamz nodes can decide what to do with it to get out any metrics of interest. On the other hand, the model is mutable, so downstreams had better not be async, else they may process on incorrect state.

Is there any obvious "output" to the fitting and current state that we could consider as a result of the partial fit?

src/skmultiflow/core/stream.py
```diff
@@ -0,0 +1,28 @@
...
+
+@Stream.register_api()
+class from_sklearn_pipeline(Stream):
+    """
+    take a pipeline and build a stream
+    """
+    pass  # TODO
```
martindurant: This would function like a generator? Strictly speaking, this looks like a Source, but since I don't suppose a user would do any processing to it before passing to a learning model (it was already processed by the pipeline), not sure there would be much use in exposing this.

src/skmultiflow/data/base_generator.py
```diff
@@ -283,20 +283,20 @@ def has_more_samples(self):
...
             A string representing the plot name.
-        
+
         """
         return self.name + " - {} target(s), {} classes, {} features".format(self.n_targets,
                                                                              self.n_classes, self.n_features)
 
     def get_class_type(self):
-        return 'stream'
+        return type(self)
```
martindurant: Always a multiflow-specific type here? Streamz comes in many classes, hope it's not possible to end up here with one of those.

src/skmultiflow/data/data_generator.py
martindurant: Unless the users are well-involved in the development process, I would second deprecation or clear documentation on how to switch.

src/skmultiflow/data/data_generator.py
```diff
@@ -0,0 +1,74 @@
...
+    def next_sample(self, batch_size=1):
+        sample = self.data.iloc[self.sample_idx:self.sample_idx + batch_size]
+        self.sample_idx += batch_size
+        if not sample.empty:
+            return sample.values if self.return_np else sample
+        else:
+            return (None, None)
+
+    def last_sample(self):
+        return self.next_sample()
```
martindurant: Is this intentional? I would not expect last sample and next sample to mean the same thing.

src/skmultiflow/data/stream.py
martindurant: `from_multiflow_generator` ?

src/skmultiflow/data/stream.py
```diff
@@ -0,0 +1,30 @@
...
+        self.stopped = False
+        self.started = True
+        self.loop.add_callback(self.do_poll)
+
+    @gen.coroutine
+    def do_poll(self):
+        while self.mf_gen.has_more_samples():
+            sample = self.mf_gen.next_sample(self.batch_size)
+            yield self._emit(sample)
+            yield gen.sleep(self.poll_interval)
```
martindurant: `if self.stopped: break` ?
A common pattern in other async sources, allowing `stream.stop()`.

Is there a practical consequence to having this async? I doesn't need to be, to get closer to the functionality and performance overhead of the previous version. I note that the downstream fit/predict functionality is not async.

src/skmultiflow/demos/_test_filters.py
```diff
@@ -1,32 +1,39 @@
...
 
-    filter = MissingValuesCleaner(-47, 'median', 10)
+    """
+    df = pd.read_csv(os.path.join(curr_dir, '../data/datasets/covtype.csv'))
+    gen = DataGenerator(df, return_np=True)
+    stream = Source.from_generator(gen, batch_size=1, poll_interval=.001)
+    sliding_array = stream.sliding_window(10).map(pd.np.concatenate)  # get last ten elements
+    data_from_stream = sliding_array.map(SimpleImputer(missing_values=-47, strategy='median').fit_transform).sink_to_list()
+    stream.start()
+    time.sleep(1)  # wait long enough to let all batches pass through the stream
```
martindurant: I would recommend against this in tests, but maybe OK in a demo. There are helpers like [wait_for](https://github.com/python-streamz/streamz/blob/master/streamz/utils_test.py#L107) that could be useful.

tests/data/test_data_generator.py
```diff
@@ -0,0 +1,54 @@
+import os
+import numpy as np
+import pandas as pd
+from skmultiflow.data import DataGenerator
+
+
+def test_data_stream_X_y(test_path, package_path):
+    test_file = os.path.join(package_path, 'src/skmultiflow/data/datasets/sea_stream.csv')
```
martindurant: Is it reasonable to split this into multiple tests?

tests/data/test_stream.py
```diff
@@ -0,0 +1,16 @@
...
+from skmultiflow.data import from_generator, DataGenerator
+
+def test_sink_to_list():
+    df = pd.DataFrame(pd.np.random.randint(0, 100,size=(300, 4)), columns=list('ABCD'))
+    gen = DataGenerator(df)
+    stream = streamz.Source.from_generator(gen, batch_size=20, poll_interval=.001)
+
+    sub_df_list = stream.sink_to_list()
+    stream.start()
+    time.sleep(1.0)  # sleep long enough to let all sub_df pass through the stream
```
martindurant: Should certainly use `wait_for` here, or (perhaps and) run this as a [coroutine](https://github.com/python-streamz/streamz/blob/master/streamz/utils_test.py#L57) and use `await_for`. You can't guarantee that CI runs as fast as your local test system, and you don't want lots of `sleep` statements to slow down the testing process. Of course, if the simple data generators are not aync at all, then you will need no wait.

tests/lazy/test_sam_knn.py
```diff
@@ -73,7 +77,7 @@ def test_sam_knn_coverage(package_path):
     wait_samples = 20
 
     while cnt < max_samples:
-        X, y = stream.next_sample()
+        X, y = X_gen.next_sample(), y_gen.next_sample()
```
martindurant: Why was it necessary to split into two what was previously a single data source?

src/skmultiflow/core/stream.py
remiadon: I am not convinced about a "more explicit name", if there is a problem then it can certainly be fixed via a different way to call the method ... 

I was thinking of something like this, which would minimize change on the existing code

```python
import functools
def stream_predict(est, stream):
     # streaming dataframe instead of stream ?
     pred_stream = Stream().map(est.predict)
     pred_stream.connect(stream)
     return pred_stream

def make_stream_estimator(est):
    predict = functools.partial(stream_predict, est=est)
    setattr(est, 'stream_predict', predict)
    return est
    # similar for partial_fit
...
...
from sklearn.linear_model import LogisticRegression  
lr = make_stream_estimator(LogisticRegression())
lr.stream_predict(some_stream).map(print)  # print out predictions
```

src/skmultiflow/data/base_generator.py
remiadon: @jacobmontiel do you have an example ? This was to avoid confusion with streamz.Stream. 
IMO the extended-classes I have seen literally are generators, but I may be confused ...

src/skmultiflow/data/base_generator.py
remiadon: @jacobmontiel agreed for this PR I'll re-add this

src/skmultiflow/data/data_generator.py
remiadon: File sources are already tackled by streamz via `Source.from_textfile`

I think I can add a deprecation warning

src/skmultiflow/core/stream.py
remiadon: Following the sklearn way of doing, the output of partial_fit should be `self.model`
But in this case we are in a different context, calls in streamz may be asynchronous, which get things complicated

Option 1 : 
Return self.model, and to make very explicit for users that this part of the code is not responsible for ensuring process on a correct state

Option 2 (requires more work):
Interally manage the access to self.model (via [tornado conditions ? ](https://www.tornadoweb.org/en/stable/locks.html))

martindurant: After some consideration, Option 1 is OK for the moment - with appropriate caveats - and I don't expect we actually expect any async down-streams here for the moment. In examples, it would be good to show what you might do with the model to get some basic metrics out for plotting the performance of the model versus time.

remiadon: @martindurant to answer to your question more deeply, most of the informations you could get from a model actually depend on the "state" of the model
Even something like a call to "get_infos()" would lead to incorrect results if runned asynchronously. So I don't see what useful information can be returned ...

src/skmultiflow/core/stream.py
remiadon: This something to drop.
My actual idea is to provide a helper to get users from the sklearn comunity
That could be a function that constructs a streamz.Stream from a sklearn.Pipeline

If you already have a pipeline, you would call something like
```python
stream = Stream.from_sklearn_pipeline(pipeline)
```

I agree there is a confusion with the term "Pipeline". A stream can actually be seen as a pipeline. That kind of helper could be used to improve interactions from both worlds (static pipelines from sklearn, streams from streamz)

This is only an idea for now. I'll remove this and save it for later

src/skmultiflow/core/stream.py
martindurant: `.map(lambda x: x.get_infos())` would run immediately, generate a string, and be correct at the moment it is called. *If* there are downstream nodes that want to do something else with the model object, they may be slightly out-of-date, but not by much. If the object is to plot a time-series of the model performance, I don't anticipate much of a problem.

src/skmultiflow/demos/_test_filters.py
remiadon: I did not know there was a function for this 
Thanks !!

tests/data/test_stream.py
remiadon: Thanks for that precious comment
I'll update this

src/skmultiflow/core/stream.py
remiadon: I think I got your point
I would actually be OK for returning get_infos() for a V1

src/skmultiflow/data/stream.py
remiadon: No I don't see any consequence of making this async

martindurant: To be checked: I think you will get exactly one event emitted per event loop tick, which may not be significant overhead compared to the time spent training for each sample.

tests/data/test_stream.py
remiadon: @martindurant actually I have some problems with `wait_for`

`sub_df_list` is empty when it's declared
What I usually do is I make a call to pd.concat on this list
calling pd.concat on the current state of the list raises an error, even using `wait_for`, which is normal since the concat is done before the waiting ...

src/skmultiflow/data/data_generator.py
remiadon: @jacobmontiel for this case I don't see how a deprecation warning could help, taking the renaming into account

If users had pieces of code importing FileStream, they could not import it anyway because following the applied renaming it would be rename to FileGenerator ...

src/skmultiflow/data/stream.py
remiadon: updated

tests/data/test_data_generator.py
remiadon: Done

tests/data/test_stream.py
remiadon: I managed to to this by creating a custom predicate and submit it to wait_for

src/skmultiflow/demos/_test_filters.py
remiadon: Actually @martindurant this is not a test, this is a demo.
In the future if demos are replaced by notebooks (I raised an issue for this)
Then calls to sleep or wait_for would not make sense
The cells of the notebook would have to be executed in time, that's all

src/skmultiflow/data/data_generator.py
remiadon: @martindurant actually the way it is implemented in mutlfilow is that generator (formerly streams) keep the last "emitted" batch as an internal attributes. 
I can change this piece to fit more precisely to the definition

martindurant: So long as it's documented :)

src/skmultiflow/data/base_generator.py
jacobmontiel: Currently, there are two types of data sources supported: generators and data batches (files and raw data).

In this case, FileStream and DataStream are not generators. Future sources (via streamz) that come to mind include Kafka, TCP which also do not fall in the generator category.

During the first stage of development, we focused on generators because they are cheap (memory-wise). However, the next step is to allow multiple data sources. Generators are nice for reproducibility, but in order to make `skmultiflow` useful for real-world applications, we need to provide the user with more options.

src/skmultiflow/data/data_generator.py
jacobmontiel: I mean to have a wrapper class `FileStream` (same name as before) that internally calls the new method. If the user calls `FileStream` then the warning is raised.

src/skmultiflow/core/stream.py
jacobmontiel: `get_info` is just a method to provide the initial configuration of the estimator, so it does not change as the model is trained. I prefer `return self` because the user could then call something like `model.score(X, y_true)` to have an estimation on performance. `score()` is now available in all classifiers and regressors.

src/skmultiflow/data/base_generator.py
jacobmontiel: `get_class_type` has been deprecated. However, we could take the chance to define a method in the base class to return the stream (streamz) type.

remiadon: > However, we could take the chance to define a method in the base class to return the stream (streamz) type

Don't think so since this PR tries to separate generators from streams

src/skmultiflow/data/data_generator.py
jacobmontiel: This is indeed misleading. Here is the current implementation in the `Stream` class:
``` python
    def last_sample(self):
        """ Retrieves last `batch_size` samples in the stream.

        Returns
        -------
        tuple or tuple list
            A numpy.ndarray of shape (batch_size, n_features) and an array-like of shape
            (batch_size, n_targets), representing the next batch_size samples.

        """
        return self.current_sample_x, self.current_sample_y
```

**Note:** The docstring should read "...representing the **last** batch_size samples."

PR #118, opened 2019-06-11
Change default window size in KNN + ADWIN classifier

foxriver76:
- sys.maxsize seems to be inappropriate
- use windowsize as in https://github.com/scikit-multiflow/scikit-multiflow/blob/19a08d268d744433a52b4812a655035e402c46f2/src/skmultiflow/lazy/knn.py#L95
- e. g. sys.maxsize ends up in a window size of 9223372036854775807 on my system and also in the documentation https://scikit-multiflow.github.io/scikit-multiflow/skmultiflow.lazy.knn_adwin.html?highlight=knn%20adwin#module-skmultiflow.lazy.knn_adwin

* 
* 
* Fixes # . [Optional]

---

Checklist

### DIFF
--- a/src/skmultiflow/lazy/knn.py
+++ b/src/skmultiflow/lazy/knn.py
@@ -26,13 +26,13 @@ class KNN(BaseSKMObject, ClassifierMixin):
     
     Parameters
     ----------
-    n_neighbors: int
+    n_neighbors: int (default=5)
         The number of nearest neighbors to search for.
         
-    max_window_size: int
+    max_window_size: int (default=1000)
         The maximum size of the window storing the last viewed samples.
         
-    leaf_size: int
+    leaf_size: int (default=30)
         The maximum number of samples that can be stored in one leaf node, 
         which determines from which point the algorithm will switch for a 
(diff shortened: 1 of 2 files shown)


jacobmontiel:
Thanks for this change @foxriver76. Could you update the docstring to explicitly state the default values, something like:

```
   Parameters
    ----------
    n_neighbors: int (dafault=5)
        The number of nearest neighbors to search for.
        
    max_window_size: int (dafault=1000)
        The maximum size of the window storing the last viewed samples.
        
    leaf_size: int (dafault=30)
        The maximum number of samples that can be stored in one leaf node, 
        which determines from which point the algorithm will switch for a 
        brute-force approach. The bigger this number the faster the tree 
        construction time, but the slower the query time will be.
```

This also applies to the parent class (`skmultiflow.lazy.KNN`)

Let me know what you think.

foxriver76:
Sorry for the late response. I've added the information to the docstring.

jacobmontiel:
No problem at all @foxriver76 

Thanks again for this change. I am merging it.

PR #119, opened 2019-06-17
Fixing division by zero and dealing with numerical imprecision on Regression and Multi-target Regression HT

smastelini:
Greetings @jacobmontiel,

I'm happy to finally be able to contribute again to scikit-multiflow. It's been a long time!

I've fixed some minor bugs when calculating the standard deviation and comparing split merits. These changes encompass both the Regression HT (R-HT) and Multi-target Regression HT (MTR-HT) algorithms.

The way the variance and standard deviation are calculated in the regression HT algorithms can lead to numerical imprecision. This fact is discussed [in this link](https://www.johndcook.com/blog/standard_deviation/). Considering that is not easy to change the HTs to add a more efficient online variance calculation algorithm, I added a simple check to verify whether the resulting variance is negative (in `trees/variance_reduction_split_criterion`).

Following, I will list the implemented changes.

* Fixed division by zero in the split attempt for both R-HT and MTR-HT (Addition of a conditional clause to check whether `best_suggestion.merit > 0`).
* Added check for negative variance values before calculating the standard deviation for the R-HT.

---

Checklist


jacobmontiel:
Hello @smastelini  and welcome back :-)

Thanks for submitting this fix. Could you elaborate on this?

> Considering that is not easy to change the HTs to add a more efficient online variance calculation algorithm, I added a simple check to verify whether the resulting variance is negative...

smastelini:
Hello @jacobmontiel,

By not easy, I'm referring to the way the trees store necessary statistics for making the splits. The link I sent in the previous message suggests a more robust and stable way of calculating variances. Nonetheless, this approach presupposes a recursive variance calculation strategy, which does not reflect the way the Hoeffding Tree algorithms work:

- Firstly, variance estimations need to be made in any splitting point (not only considering all the data)
- Moreover, the statistics stored by the new presented algorithm do not have the addictive properties of the current implementation of the Hoeffding Trees (combining partial statistics -- summing and subtracting). In fact, the strategy of storing the sum of `y` and `y^2` (as the current algorithms do) works very much alike the ClusStream algorithm (and its microclusters).

For these reasons, I made the proposed changes just to avoid generating warnings due to numeric inconsistencies. Looking at the MOA implementation, there is not seem to be any special treatment for these cases, too.

What do you think?

jacobmontiel:
Thanks for the clarification @smastelini 

I think the safety check is a straightforward solution to avoid negative values, considering the limitation that you mentioned in the context of the current implementation of HT.

PR #122, opened 2019-07-06
Feature: Implement Streaming Half-Space-Trees

mertozer94:
Streaming Half-Space-Trees (HS-Trees), a fast one-class anomaly detector for evolving data streams.

Theory and the pseudocode algorithm can be reviewed in:
S. C. Tan, K. M. Ting, and T. F. Liu, “Fast anomaly detection for streaming data,” in IJCAI Proceedings-International Joint Conference on Artificial Intelligence, 2011, vol. 22, no. 1, pp. 1511–1516.:

* An implementation of streaming Half Space Trees.
* [TODO] Test is missing on this commit. I have a dummy branch where I created some tests, but it is reading from a big csv file which was mentioned in the paper. If you have any suggestion
on how to integrate any tests on this PR I would be happy. Link for the dummy brach for testing: https://github.com/mertozer94/scikit-multiflow/tree/Silly-tests-Half-Space-Trees

---

Checklist


jacobmontiel:
I have started looking into the code. My first question is about the 3 classes (`HalfSpaceTrees`, `HalfSpaceTree` and `HalfSpaceTreeNode`) declared instead of the 2 in the MOA implementation (`HSTrees` and `HSTreeNode`). Why is this?

It seems to me that this adds complexity when reading the code and differentiates both implementations. However, if this is closer to the author's work I am ok with it.

I will leave some comments in the code. I won't cover all the classes at this point but will focus on `HalfSpaceTree` and `HalfSpaceTreeNode`

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+class HalfSpaceTrees(BaseSKMObject, ClassifierMixin):
+    """
+    Half Space Trees
+
+    Parameters
+    ----------
+
+    dimensions: int, required
```
jacobmontiel: Use `n_features` for consistency with the framework.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+
+        split_value: double, optional (default=0.0)
+            Mid point of the selected random dimension.
+            'p' in the original paper.
+
+        split_attribute: int, optional (default=-1)
+            Randomly selected attribute for the node.
+            'q' in the original paper.
+
+        depth: int, optional (default=0)
```
jacobmontiel: According to the documentation in MOA, the default value should be `1`:
> The depth of this node in the HSTree. The root node is at depth 1.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+class HalfSpaceTree:
+
+    def __init__(self, max_depth, dimensions, size_limit):
+        """
+        Half Space Tree
+
+        Parameters
+            ----------
```
jacobmontiel: Please remove extra blank spaces

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        dimensions: int, required
```
jacobmontiel: Use `n_features` for consistency with the framework.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+            'maxDepth' in the original paper.
+
+        size_limit: int, optional (default=50)
+            The minimum mass required in a node (as a fraction of the window size) to calculate the anomaly score.
+            'sizeLimit' in the original paper.
+        """
+        super().__init__()
+        self.max_depth = max_depth
+        self.dimensions = dimensions
+        self.root = HalfSpaceTreeNode()
```
jacobmontiel: Why not call `build_tree(...)` here? This way we avoid having two calls in the parent class:
 ```python
tree = HalfSpaceTree(self.depth, self.dimensions, self.size_limit)
tree.root = tree.build_tree(self.min, self.max)
```
This would require a slight change in the class declaration, however, the result is closer to the MOA implementation.

File: src/skmultiflow/anomaly_detection/half_space_trees.py [hunk 6]
```diff
@@ -0,0 +1,548 @@
...
+        self.size_limit = size_limit
+        self.is_learning_phase_on = True
+
+    def predict_proba(self, X, max_score):
```
jacobmontiel: I am not sure about why this functionality is now between `HalfSpaceTrees` and `HalfSpaceTree`. I will review the code once we focus on ``HalfSpaceTrees`, just leaving a comment for the moment.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        for i in range(r):
+            votes = self.get_votes_for_instance(X[i], max_score).copy()
+            y_proba = np.zeros(int(max(votes.keys())) + 1)
+            for key, value in votes.items():
+                y_proba[int(key)] = value
+                predictions.append(y_proba)
+        return np.array(predictions)
+
+    def get_votes_for_instance(self, X, max_score):
+        """ Get class votes for a single instance.
```
jacobmontiel: I am not sure about why this functionality is now between `HalfSpaceTrees` and `HalfSpaceTree`. I will review the code once we focus on ``HalfSpaceTrees`, just leaving a comment for the moment.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        X: numpy.ndarray of shape (1, n_features)
+            Instance attributes.
+
+        Returns
+        -------
+        anomaly_score: float
+            Anomaly score of the instance X.
+
+        """
+        anomaly_score = 0.5
```
jacobmontiel: Should this be `anomaly_score = 0.0`?

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        if node is not None:
```
jacobmontiel: I believe this check is redundant considering how the tree/nodes are constructed. However, I don't mind being cautious. Comments?

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+            if node.internal_node is not True or node.r < self.size_limit:
```
jacobmontiel: There is an error in the condition:
```python
if node.internal_node is not True or node.r <= self.size_limit
```

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        """
+        Builds a single half space tree.
+
+        Parameters
+        ----------
+        min: Array of floats
+            Minimum work range for every dimension.
+        max: Array of floats
+            Maximum work range for every dimension.
+        current_depth:
```
jacobmontiel: Please check the default value. Should it be set to `1`?

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+            Indicates the current depth of the tree.
+        Returns
+        -------
+        HalfSpaceTreeNode:
+            root of the tree.
+        """
+        if self.max_depth == current_depth:
+            return HalfSpaceTreeNode(depth=current_depth, internal_node=False)
+        else:
+            random_dimension = random.randint(0, self.dimensions - 1)
```
jacobmontiel: This name is confusing, consider `random_feature_idx` instead

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+
+        return anomaly_score
+
+    def build_tree(self, min, max, current_depth=0):
+        """
+        Builds a single half space tree.
+
+        Parameters
+        ----------
+        min: Array of floats
```
jacobmontiel: `min` and `max` names are shadowing the corresponding Python functions. Consider instead something like `min_values` and `max_values`

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        node: HalfSpaceTreeNode
+
+        """
+        if node is not None:
+            if node.r is not 0 and node.l is not 0:
+                node.r = node.l
+
+            if node.l is not 0:
+                node.l = 0
+
```
jacobmontiel: An `if` statement is missing:
```python
if node.depth < self.max_depth:
```

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,548 @@
...
+        """
+        Updates model to adapt to new data.
+
+        Parameters
+        ----------
+        node: HalfSpaceTreeNode
+
+        """
+        if node is not None:
+            if node.r is not 0 and node.l is not 0:
```
jacobmontiel: Why use this condition instead of the one in MOA?
```python
if is_reference_window:
```

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Even tough we define a threshold to decide whether the sample is anomalous or not, I believe it is more humanly to say if the score is closer to zero it's not an anomaly and if it is closer to one it is an anomaly. I would say that in this logic, the middle point should be quantified as the default value.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: I don't see anything in the paper justifying the default value to 1. Here depth, k in the paper, stands for the depth of the node in the tree. The root of a tree has depth one but, maximum anomaly score of a node is `max_score = self.window_size * pow(2.0, self.depth)`. And then this value is subtracted by one. So, given a tree with only one node in it, anomaly score for the tree with depth 1 and depth 0 will be 0.5 and 0 respectively. From my point of view, since every data is in the same place, there should be no anomaly in this case and that is why I put 0 as the default value.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Your question sums up a bit to what you have cited in your first comment:

> My first question is about the 3 classes (HalfSpaceTrees, HalfSpaceTree and HalfSpaceTreeNode) declared instead of the 2 in the MOA implementation (HSTrees and HSTreeNode). Why is this? It seems to me that this adds complexity when reading the code and differentiates both implementations. 

I was thinking that having the HalfSpaceTree class makes the readability a bit easier. Maybe I am thinking way too OOP but since it is a ensemble method, instances of the ensemble should be able to decide individually also. For my point of view, a Half Space Tree should be the candidate for those functions not a Half Space Tree Node. Maybe you can explain why you were thinking like that.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Yes I agree that the check is redundant. If the node is an internal node, then it has left and right child. Otherwise, it is calculating the score. I can remove the check or continue to be cautious ?

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Yes, I agree. I have rechecked the paper and it says fewer or equal instances.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: There is 

> if node is not None:

In the beginning of that code snippet, I don't know why we should add that if statement.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: I am citing directly from the paper:

> Update model : 
> 1. Node.r ← Node.l for every node with non-zero mass r or l
> 2. Reset Node.l ← 0 for every node with non-zero mass l

Implementation from MOA:

```
		this.r = this.l;
		this.l = 0;
		if(internalNode)
		{
			left.updateModel();
			right.updateModel();
		}
```

Looking to the algorithm provided, I don't know why we should not update leaves also.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: I want to let you know that I have integrated all the small code refactoring.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: Looking closer into this, it seems that this value is actually never used. This would only happen if `node is None` which does not seem possible (given the structure of the code.)

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: I vote to remove it, for readability of the code.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: This is not the same if condition, or am I missing something?

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: I see, thanks for the reference. Two comments:

1. From the paper, should the condition be:
```python
if node.r is not 0 or node.l is not 0:
```
2. It seems to me that both implementations are updating all nodes (including leaf nodes). In MOA, the condition `if(internalNode)` is used to stop the update, whereas your implementation has the condition `if node is not None` at the begining.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: I see, thanks for the clarification. This is indeed strange, given that the root node of a tree should have a depth of `0`. As you mention, if there is only one node we cannot define an anomaly value thus `score=0`.

However, this change has a direct impact on the performance in the method. It is important to keep it in mind when validating the results from your testing. This could be the root cause on the performance difference you reported.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: I agree with your reasoning. I assumed it what you mentioned, but asked for confirmation. :+1: for readibility.

My comment here is about instantiating a Half--space tree on just one line instead of two:

```python
tree = HalfSpaceTree(self.depth, self.dimensions, self.size_limit, self.min, self.max)
```

With a call to `build_tree` from the `__init__` method.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Yes it is true, since the node never will be `None`. I can remove it then?

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: I am a bit confused on this one. Could you a bit explain why you would like to add this `if `statement?

`if node.depth < self.max_depth:
`

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: 1. You are completely right, it should be an `or`. 

1. Yes I think I was considering another term when I was writing the comment, both implementations are updating leaves. I think I was wanted to be cautious and used `if node is not None:` instead of `if is_reference_window:`. Since we removed `if node is not None:` in the other parts of the code and I can do the same here and replace it with `if is_reference_window:` ?

jacobmontiel: ~I vote to remove it. As far as I can tell there is no reason to expect such a condition.~
Looking back into this, we need the condition to stop the recursive update (in MOA is done via the internalNode check)

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: You are right, I was confused. The `if node is not None` condition has the same purpose. My bad.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Yes, I understand now, it makes it easier to read.

mertozer94: I have integrated some changes, is it possible to resolve some of the conversation? It started to be a bit hard to navigate.

jacobmontiel:
Thanks for the changes @mertozer94 !

I have marked as "resolved" the corresponding comments. Let me know if you need help with the validation testing.

On the other hand, the errors in the Codecov Report will be addressed with functional tests once we are sure the implementation is correct.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: Actually, this was the root of the performance difference that I have reported, now, depending on the parameters, auc score is around 0.998.

mertozer94:
> Thanks for the changes @mertozer94 !
> 
> I have marked as "resolved" the corresponding comments. Let me know if you need help with the validation testing.
> 
> On the other hand, the errors in the Codecov Report will be addressed with functional tests once we are sure the implementation is correct.

My tests are complying with the results of the paper, what kind of functional tests I should focus on?

jacobmontiel:
The idea behind our tests is to protect the code against unintentional changes from future development, whereas implementation correctness is tested as part of the PR process.

Considering this, we usually favor the usage of generators given that they are requirement-wise efficient. However, if there is a case where a generator is not the best option then we can consider using a data set. The most important elements are reproducibility and coverage.

If you have installed `pytest` then you can evaluate coverage offline as follows:

```
pytest --cov=src/skmultiflow tests/anomaly_detection/test_half_space_trees.py --showlocals -v
```

---
Regarding this comment related to the depth of a `HalfSpaceTreeNode`:

> Actually, this was the root of the performance difference that I have reported, now, depending on the parameters, auc score is around 0.998.

Does this means that the default value should be set as in MOA (`depth=1`) or can it be set by the user?

jacobmontiel:
@mertozer94 It might be usefult to take a look into `/tests/trees/test_hoeffding_tree.py` for some ideas

mertozer94:
Thank you for the clarification. 
> Regarding this comment related to the depth of a `HalfSpaceTreeNode`:
> 
> > Actually, this was the root of the performance difference that I have reported, now, depending on the parameters, auc score is around 0.998.
> 
> Does this means that the default value should be set as in MOA (`depth=1`) or can it be set by the user?

I would say that we can set the default value as 1 and not leave it to the user.

jacobmontiel:
> I would say that we can set the default value as 1 and not leave it to the user.

Sounds good to me.

mertozer94:
Hello @jacobmontiel I needed a generator which was capable of generating values between 0 to 1. I found SEAGenerator the most suitable one. It creates values between 0 to 10 and I am simply scaling those into 0 to 1. I have included a really simple test case which you can find in tests/anomaly_detection/test_half_space_trees

jacobmontiel:
Thanks @mertozer94 

It seems that `codecov` is broken for the PR. This is likely related to the lack of support for "force-pushed" commits. Could you run locally the coverage report?

mertozer94:
> Thanks @mertozer94
> 
> It seems that `codecov` is broken for the PR. This is likely related to the lack of support for "force-pushed" commits. Could you run locally the coverage report?

Of course, here is the result when I run it locally:

```
/home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_tree.py                             55     55     0%
/home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_tree_node.py                        11     11     0%
/home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_trees.py                           176      3    98%
```
And in total:
`TOTAL                                                                                                                   12255   2410    80%
`

P.S: I had to remove `test_show_versions.py` because I had the following error:

```
../../.local/lib/python3.6/site-packages/skmultiflow/utils/_show_versions.py:61: ImportError
ImportError: cannot import name '_get_blas_info'
```

jacobmontiel:
Thanks @mertozer94 

If I am correct these two files are not part of your PR:

> ```
>  /home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_tree.py                             55     55     0%
>  /home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_tree_node.py                        11     11     0%
```

The PR is almost ready. I was revisiting the open actions from the review. Please check my comments. Also, I added some comments in your test.

File: src/skmultiflow/anomaly_detection/half_space_trees.py [hunk 16]
```diff
@@ -0,0 +1,557 @@
...
+        Returns
+        -------
+        numpy.array
+            Predicted the probabilities of all the labels for all instances in X.
+
+        """
+        r, _ = get_dimensions(X)
+        predictions = []
+        for i in range(r):
+            votes = self.get_votes_for_instance(X[i], max_score).copy()
```
jacobmontiel: Please use `copy.deepcopy` to protect the internal votes. We have had some issues due to `.copy()` not working as expected.

```python
            votes = copy.deepcopy(self.get_votes_for_instance(X[i], max_score))
```

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,557 @@
...
+        for i in range(r):
+            votes = self.get_votes_for_instance(X[i])
+            if votes == {}:
+                # Ensemble is empty, all classes equal, default to zero
+                predictions.append(0)
+            else:
+                predictions.append(max(votes, key=votes.get))
+        return np.asarray(predictions)
+
+    def get_votes_for_instance(self, X):
```
jacobmontiel: This name is confusing, given that the method does not only retrieves the code but also determines the actual prediction. In this sense, I would prefer to rename it as `_predict` given that is the private function, called by `predict`, that actually determines the prediction value.

```python
   def _predict(self, X):
```

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,557 @@
...
+            The features to train the model.
+
+        y: numpy.ndarray of shape (n_samples)
+            An array-like with the class labels of all samples in X.
+
+        classes: numpy.ndarray, optional (default=None)
+            Array with all possible/known class labels. This is an optional parameter, except
+            for the first partial_fit call where it is compulsory.
+
+        sample_weight: numpy.ndarray of shape (n_samples), optional (default=None)
```
jacobmontiel: This parameter is not used by this method. Please update the documentation to clearly state this fact.

Note: Do not remove the parameter given that it is defined by the `CalssifierMixin` class.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,557 @@
...
+
+        Parameters
+        ----------
+        X : numpy.ndarray of shape (n_samples, n_features)
+            The features to train the model.
+
+        y: numpy.ndarray of shape (n_samples)
+            An array-like with the class labels of all samples in X.
+
+        classes: numpy.ndarray, optional (default=None)
```
jacobmontiel: This parameter is not used by this method. Please update the documentation to clearly state this fact.

Note: Do not remove the parameter given that it is defined by the `CalssifierMixin` class.

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -0,0 +1,557 @@
...
+        -------
+        dict (class_value, weight)
+
+        """
+
+        combined_votes = {}
+        max_score = self.window_size * pow(2.0, self.depth)
+
+        for i in range(self.n_estimators):
+            vote = self.ensemble[i].get_votes_for_instance(X, max_score)
```
jacobmontiel: This line partially duplicates the functionality of `predict_proba()` and leaves the internal `votes` values exposed. My suggestion is to change this line to:

```python
            vote = self.ensemble[i].predict_proba(X)
```

**Note 1:** This way you will get an array instead of a dictionary.
**Note 2:** With this change, `max_score` in no longer required.
**Note 3:** This way the `predict` and `predict_proba` are related (as desired) and potential bugs can be easily identified if either method is impacted by future (unrelated) changes.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: This functionality is correct. No further action required.

tests/anomaly_detection/test_half_space_trees.py
```diff
@@ -0,0 +1,33 @@
...
+    y_proba = []
+    wait_samples = 500
+
+    while cnt < max_samples:
+        X, y = stream.next_sample()
+        # Scale inputs between 0 and 1
+        X = X / 10
+        if (cnt % wait_samples == 0) and (cnt != 0):
+            y_pred.append(learner.predict(X)[0])
+            y_proba.append(learner.predict_proba(X)[0])
```
jacobmontiel: This line serves for coverage. However, given its relevance, it is important to have an explicit check on the actual probability values.

Something in the lines of:

```python
assert np.allclose(y_proba, expected_proba)
```

You can explicitly write the expected probability values (`expected_proba`) or save them in a `numpy` file to be loaded by the test.

mertozer94:
> Thanks @mertozer94
> 
> If I am correct these two files are not part of your PR:
> 
> > ```
> >  /home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_tree.py                             55     55     0%
> >  /home/mert/.local/lib/python3.6/site-packages/skmultiflow/anomaly_detection/half_space_tree_node.py                        11     11     0%
> > ```
> 
> ```
> 
> The PR is almost ready. I was revisiting the open actions from the review. Please check my comments. Also, I added some comments in your test.
> ```

Yes they are not part of the the PR.

mertozer94:
> Here are my last "minor" comments. The PR is almost ready, thanks for your understanding.
> 
> **Note:** Please avoid force pushing the commits in your branch since this could corrupt the PR. [[Source](https://help.github.com/en/articles/about-pull-requests)]

No worries, since it was my first PR in a open source library I was expecting to have a lot of comments. Now I know more about the structure so I am more prepared for the next PR :)

I was squashing latest fix in order to make PR ready to merge all the time.

src/skmultiflow/anomaly_detection/half_space_trees.py
mertozer94: > This line partially duplicates the functionality of `predict_proba()` and leaves the internal `votes` values exposed. My suggestion is to change this line to:
> 
> ```python
>             vote = self.ensemble[i].predict_proba(X)
> ```
> 
> **Note 1:** This way you will get an array instead of a dictionary.
> **Note 2:** With this change, `max_score` in no longer required.
> **Note 3:** This way the `predict` and `predict_proba` are related (as desired) and potential bugs can be easily identified if either method is impacted by future (unrelated) changes.

I agree with all points. I guess you meant:
```
            vote = self.predict_proba(X)
```
Which makes use of the method `def predict_proba(self, X):` in class HalfSpaceTrees(which is not used).

tests/anomaly_detection/test_half_space_trees.py
mertozer94: > This line serves for coverage. However, given its relevance, it is important to have an explicit check on the actual probability values.
> 
> Something in the lines of:
> 
> ```python
> assert np.allclose(y_proba, expected_proba)
> ```
> 
> You can explicitly write the expected probability values (`expected_proba`) or save them in a `numpy` file to be loaded by the test.

I have included the change, however probabilities are a bit different and test is not passing every time. I will try to find the root of this problem.
For now, I am just sharing the result:

```
expected_proba = array([[7.81250000e-06, 9.99992187e-01],
       [1.61760000e-01, 8.38240000e-01],
       [1.98240000e-01, 8.01760000e-01],
       [2.19360000e-01, 7.80640000e-01],
       [1.72000000e-01, 8.28000000e-01],
       [7.29837500e-02, 9.27016250e-01],
       [2.16320000e-01, 7.83680000e-01],
       [2.13920000e-01, 7.86080000e-01],
       [2.07373125e-01, 7.92626875e-01]])
```
```
y_proba    = [array([7.81250000e-06, 9.99992187e-01]),
 array([0.14832, 0.85168]),
 array([0.18864, 0.81136]),
 array([0.20784, 0.79216]),
 array([0.16384, 0.83616]),
 array([0.07984, 0.92016]),
 array([0.20784, 0.79216]),
 array([0.20464, 0.79536]),
 array([0.21568, 0.78432])]
```

mertozer94: > > This line serves for coverage. However, given its relevance, it is important to have an explicit check on the actual probability values.
> > Something in the lines of:
> > ```python
> > assert np.allclose(y_proba, expected_proba)
> > ```
> > 
> > 
> > You can explicitly write the expected probability values (`expected_proba`) or save them in a `numpy` file to be loaded by the test.
> 
> I have included the change, however probabilities are a bit different and test is not passing every time. I will try to find the root of this problem.
> For now, I am just sharing the result:
> 
> ```
> expected_proba = array([[7.81250000e-06, 9.99992187e-01],
>        [1.61760000e-01, 8.38240000e-01],
>        [1.98240000e-01, 8.01760000e-01],
>        [2.19360000e-01, 7.80640000e-01],
>        [1.72000000e-01, 8.28000000e-01],
>        [7.29837500e-02, 9.27016250e-01],
>        [2.16320000e-01, 7.83680000e-01],
>        [2.13920000e-01, 7.86080000e-01],
>        [2.07373125e-01, 7.92626875e-01]])
> ```
> 
> ```
> y_proba    = [array([7.81250000e-06, 9.99992187e-01]),
>  array([0.14832, 0.85168]),
>  array([0.18864, 0.81136]),
>  array([0.20784, 0.79216]),
>  array([0.16384, 0.83616]),
>  array([0.07984, 0.92016]),
>  array([0.20784, 0.79216]),
>  array([0.20464, 0.79536]),
>  array([0.21568, 0.78432])]
> ```

So, I found the root of the problem, which was basically creation of trees before setting the `random_state`. I also added `from skmultiflow.utils import check_random_state` instead of setting random seed manualy in the code.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: Using `self.predict_proba` is another option. 

Notice that in both cases, you could remove completely this method and move the functionality to `self.predict`. The thing is that `self.predict_proba` and `self.ensemble[i].predict_proba(X)` support multiple samples in `X` whereas this function process one sample at a time (since it is called inside a `for` loop).

I am in favor of using `self.predict_proba` since it will make the code more readable.

tests/anomaly_detection/test_half_space_trees.py
jacobmontiel: Awesome! Just in case, you can look here `src/skmultiflow/meta/classifier_chains.py` for an example on the standard way we use `random_state`.

Notice in the `__init__` method that there are two parameters:

```python
        self.random_state = random_state    # This parameters has the same name as the parameter in the method's signature
        self._random_state = None           # This is the actual random_state object used internally
```

`self.random_state` is used by the `get_info` method and `reset` methods and it only keeps the **value** that was passed by the user. `self._random_state` is the actual `random` **object** used internally. It is important to make sure that you **only** use `self._random_state` internally.

mertozer94:
@jacobmontiel I have pushed the requested changes.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: See comment below.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: Please restore the parameter in the docstring since it is used to generate the documentation. What I mean to say is that it should be stated that the parameter is in the signature but not used, sorry for not being clear.

Suggestion:
```
        classes: None
            Not used by this method.
        sample_weight: None
            Not used by this method.
```

src/skmultiflow/anomaly_detection/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+The :mod:`skmultiflow.anomaly_detection` module includes anomaly detection methods based.
+"""
```
jacobmontiel: There is a typo, suggested text:

```
Please restore the parameter in the docstring since it is used to generate the documentation. What I mean to say is that it should be stated that the parameter is in the signature but not used, sorry for not being clear.

Suggestion:
```
The :mod:`skmultiflow.anomaly_detection` module includes anomaly detection methods.
```

mertozer94: > Thanks for the last changes. The code seems ready, there are just minor documentation changes left that I highlighted.
> 
> Please use a normal commit (not a forced one) to upload the new versions. The reason for using a normal commit is to try to get `codecov` to recognize the change and provide the coverage report for the entire project, not only for your changes ;-)

I will keep that in mind :) By the way, I have added the documentation changes.

jacobmontiel:
Awesome work @mertozer94 and thanks for your patience during the PR review.

All changes have been implemented. The PR is ready to merge.

mertozer94:
> thanks for your patience during the PR review.

@jacobmontiel It's a pleasure :) I feel more comfortable with the environment right now and I would like to continue to implement anomaly detection algorithms, if it is possible of course.

jacobmontiel:
@mertozer94, of course. We are open to contributions and collaborations. If you already have an anomaly detection method in mind please create a new issue and we can discuss it. We will look into this on our side as well.

PR #129, opened 2019-08-08
Bug fix for hyperplane generator

AnhTruong:
Currently, there's some issues on:
- calculation of sum of weights
- calculation of sum of weights*features
Those get accumulated over all samples, then the results look random: we cannot see the change of hyperplanes.

In this PR, I've added some code to:
- modify those calculations
- add a test to verify the change
- modify the pre-stored data saved in hyper_plane_stream.npz

All tests passed.


AnhTruong:
Hi @jacobmontiel , I've added a PR to fix the bug with hyperplane generator. Please check. Thank you.

AnhTruong:
Part of the test uses `batch_size=10` when the stream generates 10 samples and the test then validates both features and targets. It fails at the last target out of ten. The followings would be self-explanatory:

` stream.restart()
        X, y = stream.next_sample(10)
        assert np.alltrue(X == X_expected)
>       assert np.alltrue(y == y_expected)
E       assert False
E        +  where False = <function alltrue at 0x000002786C8438C8>(array([1, 1, ..., 0, 1, 0, 0]) == array([1., 1.,..., 1., 0., 1.])
E        +    where <function alltrue at 0x000002786C8438C8> = np.alltrue
E           -array([1, 1, 1, 0, 1, 1, 0, 1, 0, 0])
E           +array([1., 1., 1., 0., 1., 1., 0., 1., 0., 1.])
E           Full diff:
E           - array([1, 1, 1, 0, 1, 1, 0, 1, 0, 0])
E           ?                                   ^
E           + array([1., 1., 1., 0., 1., 1., 0., 1., 0., 1.])
E           ?         +   +   +   +   +   +   +   +   +  ^^)`

jacobmontiel:
I missed that check which indded is related to your fix. Thanks for the clarification.

jacobmontiel:
Commit c1704a9dc10e2288317464979b8ea556164bbc2f complements this fix to account for the cases when drift is added to the stream.

PR #134, opened 2019-09-01
Duplicated lines in prequential documentation example

luccaportes:
Hi, I realized there are two duplicated calls to `evaluate` in two parts of the documentation of the Prequential Evaluator. This PR simply removes them.

*  Removed duplicated lines in the Prequential evaluator

---

I think the checklist doesn't apply to this pull request, since there is no new code proposed.


jacobmontiel:
@luccaportes Thanks for fixing this!

PR #137, opened 2019-09-12
Fix bug in precision, recall, F1 and geometric mean scores

jacobmontiel:

During work to improve test coverage of the evaluators a bug was found in the calculation of precision and recall which in places affects the f1-score and geometric mean scores.

The bug is triggered by the order of the observed classes. This is because of the confusion matrix used internally which maps the observed classes by the order of arrival and not the class-value itself.

PR is composed of:
1) Fix bug in precision, recall, f1-score, and geometric mean
2) Improved tests for metrics
3) Improved test coverage for evaluators

---

Checklist


smastelini:
I think this PR is ready to merge!

PR #143, opened 2019-09-29
Fix index out of range for the classification functions in the mixed …

ingako:
The mixed generator only has 2 classification functions [0, 1].

* Fixes the index out of range error when generating drifts by the mixed generator.

---

Checklist

- [X] Travis CI build passes with no errors.
- [X] Test Coverage is maintained (threshold is -0.2%).

### DIFF
--- a/src/skmultiflow/data/mixed_generator.py
+++ b/src/skmultiflow/data/mixed_generator.py
@@ -286,10 +286,7 @@ class MIXEDGenerator(Stream):
 
     def generate_drift(self):
         """
-        Generate drift by switching the classification function randomly.
+        Generate drift by switching the classification function.
 
         """
-        new_function = self._random_state.randint(2)
-        while new_function == self.classification_function:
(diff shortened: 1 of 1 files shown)


jacobmontiel:
Hello @ingako 

Thanks for the PR. Could you please update the documentation as well? There are some parts in the docstring where it is not clear that there are only two options available.

For example:

```
    classification_function: int (default: 0)
        Which of the four classification functions to use for the generation.
        Valid options are 0 or 1.
```

Also, the `mixed_geenerator` method can be simplified:

```python
    def generate_drift(self):
        """
        Generate drift by switching the classification function.

        """

        self.classification_function = 1 - self.classification_function
```

Note: I also updated the documentation for this method.

ingako:
Thanks for the comments @jacobmontiel ! 
I have pushed all your requested changes.

jacobmontiel:
Thanks @ingako !

PR is ready to merge.

Issue #131, opened 2019-08-21
[FEATURE] Adaptive Random Forest Regressor


jacobmontiel:
The `Adaptive Random Forest Regressor` is an adaptation of the `Adaptive Random Forest` classifier.

The implementation of this method is somewhat straightforward given that the original (classification) version and the base learner (Fastand Incremental Model Trees) are already implemented in `scikit-multiflow`. 

---
**Refrences**
Gomes, H.M., Bifet, A., Read, J., Barddal, J.P., Enembreck, F., Pfharinger, B., Holmes, G. and Abdessalem, T., 2017. Adaptive random forests for evolving data stream classification. Machine Learning, 106(9-10), pp.1469-1495. ([paper](https://www.elen.ucl.ac.be/Proceedings/esann/esannpdf/es2018-183.pdf))

jacobmontiel:
@AlbandeCrevoisier will be working on this. Pining @smastelini for support.

hmgomes:
There is an [ARF regressor](https://www.researchgate.net/publication/329397374_Adaptive_random_forests_for_data_stream_regression) implemented in MOA which can be used for reference (maybe comparisons?). The main difference is that it uses FIMT-DD as the base learner.

jacobmontiel:
@AlbandeCrevoisier Please create a PR when you have an initial version of the implementation so we can provide feedback.

AlbandeCrevoisier:
I — finally — opened a PR, looking for feedback :)

smastelini:
Closing since it was addressed in #145 and #237.

PR #145, opened 2019-10-03
Arf regressor

AlbandeCrevoisier:

* Adding a regressor version of Adaptive random forest.
* Renaming AdaptiveRandomForest in AdaptiveRandomForestClassifier for consistency.

---

Checklist

[Issue 131](https://github.com/scikit-multiflow/scikit-multiflow/issues/131)

I panicked & got stuck with the split file model, so eventually I backtracked to the simpler solution of copying the current ARF & modifying what needs be. I tried to document it as best I could, but might have missed something.

The testing is minimal, I don't really know how to write tests. Any suggestion?


jacobmontiel:
Some comments:

1. We need to let the users know that the ARF classifier is being renamed and must keep the old name for a couple de releases. I will be working on this next week for other methods. I will skip ARF so the renaming can be part of this PR.
2. Extending the classifier method is OK. Later we can revisit the option of streamlining the implementation for both classifier and regressor.
3. @smastelini will be providing feedback for the code.
4. @hmgomes and I can focus on the implementation's correctness. For this, we need some tests to make sure that ARFR is working as expected. Any recommendations on the data to use for this purpose @hmgomes?
5. Once correctness is assessed, we can move to functional tests.

hmgomes:
> Some comments:
> 
> 1. We need to let the users know that the ARF classifier is being renamed and must keep the old name for a couple de releases. I will be working on this next week for other methods. I will skip ARF so the renaming can be part of this PR.
> 2. Extending the classifier method is OK. Later we can revisit the option of streamlining the implementation for both classifier and regressor.
> 3. @smastelini will be providing feedback for the code.
> 4. @hmgomes and I can focus on the implementation's correctness. For this, we need some tests to make sure that ARFR is working as expected. Any recommendations on the data to use for this purpose @hmgomes?
> 5. Once correctness is assessed, we can move to functional tests.

We can use the same datasets as in this paper: https://www.researchgate.net/publication/329397374_Adaptive_random_forests_for_data_stream_regression

The results might differ a bit as in that paper the base learner was FIMTDD

Cheers,
Heitor

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+
+    max_byte_size: int, optional (default=33554432)
+        (`RegressionHoeffdingTree` parameter)
+        Maximum memory consumed by the tree.
+
+    memory_estimate_period: int, optional (default=2000000)
+        (`RegressionHoeffdingTree` parameter)
+        Number of instances between memory consumption checks.
+
+    grace_period: int, optional (default=50)
```
smastelini: Throughout the framework the standard value for this parameter is 200.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        (`RegressionHoeffdingTree` parameter)
+        Number of instances a leaf should observe between split attempts.
+
+    split_confidence: float, optional (default=0.01)
```
smastelini: The common value for this parameter is 0.0000001 (please, refer for the RegressionHoeffdingTree implementation).

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: Throughout the framework the standard value for this parameter is 200.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        (`RegressionHoeffdingTree` parameter)
+        Allowed error in split decision, a value closer to 0 takes longer
+        to decide.
+
+    tie_threshold: float, optional (default=0.05)
+        (`RegressionHoeffdingTree` parameter)
+        Threshold below which a split will be forced to break ties.
+
+    binary_split: bool, optional (default=False)
```
smastelini: The commonly used standard value for this parameter is `True`.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+
+    remove_poor_atts: bool, optional (default=False)
+        (`RegressionHoeffdingTree` parameter)
+        If True, disable poor attributes.
+
+    no_preprune: bool, optional (default=False)
+        (`RegressionHoeffdingTree` parameter)
+        If True, disable pre-pruning.
+
+    leaf_prediction: string, optional (default='nba')
```
smastelini: This should be updated to the regression case. Please, refer to the RegressionHoeffdingTree implementation for details.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        (`RegressionHoeffdingTree` parameter)
+        Prediction mechanism used at leafs.
+
+        - 'mc' - Majority Class
+        - 'nb' - Naive Bayes
+        - 'nba' - Naive Bayes Adaptive
+
+    nb_threshold: int, optional (default=0)
```
smastelini: We could add an observation here telling that this parameter does not apply in the regression case. What do you think @jacobmontiel? We could also extend this to the RegressionHoeffdingTree.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        (`RegressionHoeffdingTree` parameter)
+        The learning rate of the perceptron.
+
+    learning_ratio_decay: float
+        (`RegressionHoeffdingTree` parameter)
+        Decay multiplier for the learning rate of the perceptron
+
+    learning_ratio_const: Bool
+        (`RegressionHoeffdingTree` parameter)
+        If False the learning ratio will decay with the number of examples seen
```
smastelini: The default values should appear. Could you add them?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+    It also allows training background trees, which start training if a warning
+    is detected and replace the active tree if the warning escalates to a drift.
+
+    References
+    ----------
+    .. [1] Gomes, H.M., Bifet, A., Read, J., Barddal, J.P., Enembreck, F.,
+       Pfharinger, B., Holmes, G. and Abdessalem, T., 2017.
+       Adaptive random forests for evolving data stream classification.
+       Machine Learning, 106(9-10), pp.1469-1495. 
+    """
```
smastelini: I think the reference should be updated/extended for the ESANN 2018 paper.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+    def partial_fit(self, X, y, sample_weight=None):
+        """ Partially (incrementally) fit the model.
+
+        Parameters
+        ----------
+        X : numpy.ndarray of shape (n_samples, n_features)
+            The features to train the model.
+
+        y: numpy.ndarray of shape (n_samples)
+            An array-like with the class labels of all samples in X.
```
smastelini: This should be updated to the regression case.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        r, _ = get_dimensions(X)
+        predictions = []
+        for i in range(r):
+            votes = self.get_votes_for_instance(X[i])
+            if votes == {}:
+                # Ensemble is empty, all classes equal, default to zero
+                predictions.append(0)
+            else:
+                predictions.append(max(votes, key=votes.get))
+        return np.asarray(predictions)
```
smastelini: I believe that in the regression case an average aggregation function should be used.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        if self.max_features == 'auto' or self.max_features == 'sqrt':
+            self.max_features = round(math.sqrt(n))
+        elif self.max_features == 'log2':
+            self.max_features = round(math.log2(n))
+        elif isinstance(self.max_features, int):
+            # Consider 'max_features' features at each split.
+            pass
+        elif isinstance(self.max_features, float):
+            # Consider 'max_features' as a percentage
+            self.max_features = int(self.max_features * n)
```
smastelini: We should also check whether `max_features` is in the range `(0.0, 1.0]`.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,497 @@
...
+        self.evaluator = self.evaluator_method()
+
+    def partial_fit(self, X, y, sample_weight, instances_seen):
+        self.estimator.partial_fit(X, y, sample_weight=sample_weight)
+
+        if self.background_learner:
+            self.background_learner.estimator.partial_fit(X, y, sample_weight=sample_weight)
+
+        if self._use_drift_detector and not self.is_background_learner:
+            correctly_classifies = self.estimator.predict(X) == y
```
smastelini: This could be misleading for regression tasks. Would an acceptable amount of error be preferred?

tests/meta/test_adaptive_random_forest_regressor.py
```diff
@@ -0,0 +1,52 @@
...
+        # Test every n samples
+        if (cnt % wait_samples == 0) and (cnt != 0):
+            y_pred.append(learner.predict(X)[0])
+            y_true.append(y[0])
+        learner.partial_fit(X, y)
+        cnt += 1
+
+    error = mean_absolute_error(y_true, y_pred)
+    expected_error = 155.5787786499754
+    assert (np.abs(expected_error - error) < 1)
```
smastelini: I think that np.isclose would be better in this case.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
hmgomes: For the adaptive random forest algorithm a grace period of 50 is advisable. 
There is no mention of what value was used in that ARFReg paper, still, a good option in this case is to rely on the original ARF algorithm for classification for these parameters. 

Gomes, H. M., Bifet, A., Read, J., ..., Pfahringer, B., Holmes, G., & Abdessalem, T. (2017). Adaptive random forests for evolving data stream classification. Machine Learning, 106(9-10), 1469-1495.
*"By using smaller values of GP run time (and memory usage) is increased, and also causes trees to grow deeper, which enhances the overall variability of the forest, and consequently ARF’s classification performance. For consistency, we use the same base learner configuration for all ensembles, i.e. HNBTs with **GP = 50**."*

jacobmontiel: For consistency, we should use the same default parameters as in MOA. If there are inconsistencies with the values in the paper then a note should be added.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
jacobmontiel: Agree for the Adaptive Random Forest Regressor implementation. In the case of `RegressionHoeffdingTree` currently, there are two options available `perceptron` and `mean`.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
hmgomes: It is the default in MOA: 
https://github.com/Waikato/moa/blob/67699ac060d492986b30f888a7ffa34ca2eacf3f/moa/src/main/java/moa/classifiers/meta/AdaptiveRandomForest.java#L96

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: I think I marked the wrong line. I refer to the `nb_threshold`. Could we add some note in the documentation?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: MOA uses c=0.01,  I'd rather stick with its default values, especially reading the comments below on the `grace_period`.
I could use the value in RegressionHoeffdingTree if you think it's better.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Both RegressionHoeffdingTree & AdaptiveRandomForestClassifier use `False` as a default though, so I stuck with that. I think it's reasonable.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: Now I see your and @hmgomes points. It's better to follow the original paper! I will resolve this point! Thank your @AlbandeCrevoisier for the response.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: @hmgomes What should I do here? Only use the ESANN 2018 reference, or both?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: Ok. My bad, I verified the Regression Hoeffding Tree implementation and you are right. I'm also going to resolve this point.

tests/meta/test_adaptive_random_forest_regressor.py
AlbandeCrevoisier: I had some trouble with that: the result seemed to vary quite a lot depending on the version of python & the OS - `the tests/utils/test_utils.py::test_calculate_object_size` test fail on my machines, too. Thus, I wasn't sure which tolerance to put in the `np.isclose` test. Looking for advice on that one!

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: What should we do with that one? Mark them as not used in the documentation, & maybe issue a warning if it is set by the user?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Could you check that my fix is correct please?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: I added the values from RegressionHoeffdingTree, could you quickly check please?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Done, could you check & resolve?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
hmgomes: > @hmgomes What should I do here? Only use the ESANN 2018 reference, or both?

Use both. The ESANN one is a short paper and some details are missing from it (e.g., some of the parameters are not specified).

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: It's correct now!

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Indeed, I don't know how I could have overlooked this! It now correctly uses average aggregation instead.
@jacobmontiel , to do this I replaced get_votes_for_instance by _predict, in order to use the barycenter in the case of weighted_vote. Is that ok?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Ok, I added it, but I'm not sure of the format of the citation, could you please check it?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Done

src/skmultiflow/meta/adaptive_random_forest_regressor.py
jacobmontiel: Since this is not relevant to the regression case I vote to omit it completely from the signature and the documentation.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
jacobmontiel: The change is available in later commits:  https://github.com/scikit-multiflow/scikit-multiflow/blob/c73c759b559a76f66dfc95c9997336af712b4adb/src/skmultiflow/meta/adaptive_random_forest_regressor.py#L275-L292

Using `_predict` seems a good approach, although I have some comments:
- Renaming `votes` to something like `predicted_value` would make things even clearer.
- The MOA implementation does not take into account the performance to `weight` the predicted values. Comments @hmgomes?

src/skmultiflow/meta/adaptive_random_forest_regressor.py
jacobmontiel: This is not correct. The prediction should be used directly as input for `ADWIN`. This is also relevant since not all drift detection methods work on continuous data.

tests/meta/test_adaptive_random_forest_regressor.py
jacobmontiel: There is an issue for `test_calculate_object_size`. However, this should not affect the performance of the regressor. In the past, we had issues with the internal random number generators not being correctly set which caused errors during testing.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: I agree with that solution!

AlbandeCrevoisier: Ok, I cleaned that up.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: * Indeed, & that's now done.
* It just felt the natural way of weighting the predictions, but if it's not required we can remove the `disable_weighted_votes` parameter altogether, then.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
AlbandeCrevoisier: Indeed, I did not really think about it. I now just add the predicted value, as suggested.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: Everything seems fine to me! I'm going to resolve this comment. Thank Alban!

src/skmultiflow/meta/adaptive_random_forest_regressor.py
smastelini: Could you check that @hmgomes and @jacobmontiel?

tests/meta/test_adaptive_random_forest_regressor.py
jacobmontiel: I have committed a fix for the test 74c1bed3dc4c3af15c97e723ab57708de8ee4434 given that it has been identified to affect macOS primarily.

src/skmultiflow/meta/adaptive_random_forest_regressor.py
jacobmontiel: It seems correct to me now. Thanks!

tests/meta/test_adaptive_random_forest_regressor.py
AlbandeCrevoisier: house16H_5000-datapoints

This is what I got when training on 5000 data points of the house16H.arff dataset. The most notable difference is that error spikes, as if each trees error was the same & it was all added up. (I'll clean up the test & put it here asap.)
The MAE is comparable to that of a tree whilst no such error spike is encountered.
I'll compare with MOA's implem tomorrow to look for any functionnal difference that I missed.
I still have NaN errors, probably because of almost null sd that make NaN when normalising.

Also, test_calculate_object_size still fails here (Arch Linux).

jacobmontiel: Thanks for the information. One thing to take into account is that the random number seed is properly being propagated. Otherwise, you will get different ensembles on each run and will be unable to test the object size calculation.

AlbandeCrevoisier: house16H_5000-datapoints_nestimators1
This is with n_estimators=1... Time to debug, I guess.

Besides, I pass `self.random_state` to the trees constructor, which seems correct.

AlbandeCrevoisier: Interesting: when normalising, the variance is frequently close to 0. I haven't been able to reproduce this with the tree alone, so I assume it stems from the poisson weights of the forest. It creates some weird behaviour because we do not always check that case.

AlbandeCrevoisier: Note: ADWIN also complains *a lot* about invalid values in sqrt.

smastelini: Hello to everyone! Some valid points to consider are the differences in the strategies for normalizing features and targets used in the tree algorithms.

FIMT-DD centers the observations by the average feature value and scales them by three times the standard deviation. iSOUP-Tree and its variants use the standard z-score approach: centering by the mean and scaling by the standard deviation. The argument for the latter approach, as stated in the original proposal (here I quote Osojnik's thesis), is that:

"Ikonomovska, Gama, and Džeroski (2011b) suggest a normalization in which the difference of the value and the mean is instead divided with 3σ. However, this has the exact effect as dividing the initial learning rate by 3. Therefore, modifying the normalization procedure produces the same results as modifying the learning rate. Thus, we use the standard normalization procedure and allow a user to modify the learning rate, which already appears as a parameter of the iSOUP-Tree model." 

*Structured Output Prediction on Data Streams*. Aljaž Osojnik. PhD Thesis. Jožef Stefan International Postgraduate School. Available [here](http://kt.ijs.si/theses/phd_aljaz_osojnik.pdf).

In the case where the standard deviation is smaller than one, multiplying it by three and using the result in the denominator of the normalization might result in big numbers.

AlbandeCrevoisier: To have a baseline comparison, I ran my tests with a forest with 'mean' & no ADWIN. The results are as expected. So now, it's about 1/ 'perceptron' (surely the normalization correction will help tremendously in avoiding big numbers aberrations) 2/ ADWIN.

house16H_5000-50-mean-no-ADWIN

AlbandeCrevoisier: Update: the 'perceptron' version works fine without ADWIN...

smastelini:
Closing this PR since the work initiated here was continued in #237.

Issue #148, opened 2019-10-26
[FEATURE] Streamline Hoeffding Tree implementations


smastelini:
Currently, the Hoeffding Trees implement their leaf predictors as inner classes. This can become confusing and difficult to maintain in some cases. Besides, this kind of design decreases modularity.

Furthermore, attribute observers, split criteria, and attribute tests could also be grouped in subfolders of the skmultiflow.tree module to increase modularity.

The mentioned aspects ought to be implemented as a first step. After that, a possible direction is to verify whether there is redundant code among the leaf node predictors. In the affirmative case, we could rearrange the leaf node hierarchies to increase modularity and avoid unnecessary repetitions.

Lastly, this kind of action could also be extended to the skmultiflow.rules module in the future.

smastelini:
Addressed by PR #149.

PR #151, opened 2019-10-31
Fix of bug in IntraClusterVarianceReduction calculation

smastelini:
Fixing bug in `IntraClusterVarianceReduction` (`skmultiflow.trees`) calculation.

The current implementation of this split criterion was using `np.sum` to aggregate the `VarianceReduction` values of each target. This PR changes it to `np.mean`, as originally proposed.

* Renaming some variables for conciseness 
* Fixing the way in which `IntraClusterVarianceReduction` is calculated.

---

Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
     then pip3 install --only-binary=numpy,scipy numpy scipy;
     else pip3 install numpy scipy;
     fi
-  - pip3 install codecov pytest-cov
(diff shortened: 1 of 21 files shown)


smastelini:
Hello @jacobmontiel, this PR addresses the aspect you mentioned in #148.

Since this change does not fit within the context of the other PR, I decided to submit it separately. After that I will proceed to the another PR.

Thank you again for pointing it out.

PR #159, opened 2019-11-21
[BUG] Fix in subtree_depth calculation

smastelini:
This PR fixes a bug in the depth calculation for Hoeffding Trees.

* Fix the depth calculation for Hoeffding Trees: when iterating over their nodes, the HTs were accessing the key values of their internal structures rather than the actual nodes.
* The bug and proposed fix have no impact on the performance of the Hoeffding Tree. This method is mainly used for informative purposes via the `get_model_measurements` method.
---

Checklist


smastelini:
Hello @jacobmontiel. I have just added a test that encompasses both the `get_model_measurements` property and the `get_model_description` method.

PR #161, opened 2019-11-23
Hddm

jiahy0825:

* Add two drift detectors: HDDM_A and HDDM_W.
* Add corresponding test files.

---

Checklist

- [X] Tests are included for new functionality or updated accordingly.

---

The idea of HDDM_A and HDDM_W is based on the paper : "Online and non-parametric drift detection methods based on Hoeffding’s bounds". Implementation based on MOA.


src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+    ----------
+    drift_confidence : float (default=0.001)
+        Confidence to the drift
+
+    warning_confidence : float (default=0.005)
+        Confidence to the warning
+
+    two_side_option : bool (default=True)
+        Option to monitor error increments and decrements (two-sided) or only increments (one-sided)
+
```
jacobmontiel: Missing documentation for `lambda_option`

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+        Parameters
+        ----------
+        prediction: int (either 0 or 1)
+            This parameter indicates whether the last sample analyzed was
+            correctly classified or not. 1 indicates an error (miss-classification).
+
+        Returns
+        -------
+        HDDM_W
+            self
```
jacobmontiel: The function has no `return` statement

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def reset(self):
+        """ reset
+
+        Resets the change detector parameters.
+
+        Returns
+        -------
+        HDDM_W
+            self
```
jacobmontiel: The function has no `return` statement

src/skmultiflow/drift_detection/hddm_a.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def reset(self):
+        """ reset
+
+        Resets the change detector parameters.
+
+        Returns
+        -------
+        HDDM_A
+            self
```
jacobmontiel: The function has no `return` statement

src/skmultiflow/drift_detection/hddm_a.py
```diff
@@ -0,0 +1,193 @@
...
+        self.n_max = 0
+        self.c_max = 0
+        self.n_estimation = 0
+        self.c_estimation = 0
+
+        self.drift_confidence = drift_confidence
+        self.warning_confidence = warning_confidence
+        self.two_side_option = two_side_option
+
+    def add_element(self, prediction):
```
jacobmontiel: The function has no `return` statement

src/skmultiflow/drift_detection/hddm_a.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if self.two_side_option and self.mean_decr(self.c_max, self.n_max, self.total_c, self.total_n) :
+            self.n_estimation = self.total_n - self.n_max
+            self.c_estimation = self.total_c - self.c_max
+            self.n_min = self.n_max = self.total_n = 0
+            self.c_min = self.c_max = self.total_c = 0
+
+        self.update_estimations()
+
+    def mean_incr(self, c_min, n_min, total_c, total_n, confidence):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_a.py
```diff
@@ -0,0 +1,193 @@
...
+        self.n_min = 0
+        self.c_min = 0
+        self.total_n = 0
+        self.total_c = 0
+        self.n_max = 0
+        self.c_max = 0
+        self.c_estimation = 0
+        self.n_estimation = 0
+
+    def update_estimations(self):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_a.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def mean_incr(self, c_min, n_min, total_c, total_n, confidence):
+        if n_min == total_n:
+            return False
+
+        m = (total_n - n_min) / n_min * (1.0 / total_n)
+        cota = sqrt(m / 2 * log(2.0 / confidence))
+        return total_c / total_n - c_min / n_min >= cota
+
+    def mean_decr(self, c_max, n_max, total_c, total_n):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+                self.sample2_incr_monitor.EWMA_estimator = value
+                self.sample2_incr_monitor.independent_bounded_condition_sum = 1
+            else:
+                self.sample2_incr_monitor.EWMA_estimator = \
+                    self.lambda_option * value + aux_decay * self.sample2_incr_monitor.EWMA_estimator
+                self.sample2_incr_monitor.independent_bounded_condition_sum = \
+                    self.lambda_option * self.lambda_option + \
+                    aux_decay * aux_decay * self.sample2_incr_monitor.independent_bounded_condition_sum
+
+    def update_decr_statistics(self, value, confidence):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+                      + sample2.independent_bounded_condition_sum) * log(1 / confidence) / 2)
+        return sample2.EWMA_estimator - sample1.EWMA_estimator > bound
+
+    def monitor_mean_incr(self, confidence):
+        return self.detect_mean_increment(self.sample1_incr_monitor, self.sample2_incr_monitor, confidence)
+
+    def monitor_mean_decr(self, confidence):
+        return self.detect_mean_increment(self.sample2_decr_monitor, self.sample1_decr_monitor, confidence)
+
+    def update_incr_statistics(self, value, confidence):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+            return False
+
+        bound = sqrt((sample1.independent_bounded_condition_sum
+                      + sample2.independent_bounded_condition_sum) * log(1 / confidence) / 2)
+        return sample2.EWMA_estimator - sample1.EWMA_estimator > bound
+
+    def monitor_mean_incr(self, confidence):
+        return self.detect_mean_increment(self.sample1_incr_monitor, self.sample2_incr_monitor, confidence)
+
+    def monitor_mean_decr(self, confidence):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def detect_mean_increment(self, sample1, sample2, confidence):
+        if sample1.EWMA_estimator < 0 or sample2.EWMA_estimator < 0:
+            return False
+
+        bound = sqrt((sample1.independent_bounded_condition_sum
+                      + sample2.independent_bounded_condition_sum) * log(1 / confidence) / 2)
+        return sample2.EWMA_estimator - sample1.EWMA_estimator > bound
+
+    def monitor_mean_incr(self, confidence):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -0,0 +1,208 @@
...
+        else:
+            self.in_concept_change = False
+            self.in_warning_zone = False
+
+        self.update_decr_statistics(prediction, self.drift_confidence)
+        if self.two_side_option and self.monitor_mean_decr(self.drift_confidence):
+            self.reset()
+        self.estimation = self.total.EWMA_estimator
+
+    def detect_mean_increment(self, sample1, sample2, confidence):
```
jacobmontiel: Please make this method "non_public" using underscore as a prefix.

jiahy0825:
Thanks for your comments @jacobmontiel 

I have made changes as you requested. 

In my first PR, I have added corresponding test files and its results seems good. 
If you need further evidence, maybe I can do some comparative tests between this python code and MOA later.

jacobmontiel:
Thanks for the changes @jiahy0825 

I have some comments, which can be divided into three parts:

---
1. There are some minor items still unresolved from the review regarding some methods where the documentation states that they return `self`, however, this is not the case.

---
2. Regarding the comparison with MOA, I did a quick test using the synthetic data in your tests:

|        |  MOA | scikit-multiflow |
|:------:|:----:|:----------------:|
| HDDM_A | 1054 |       1052       |
| HDDM_W | 1020 |       1018       |

As you can see, a single drift is detected as expected and the instance ids are close and consistent `delta=2`

I did another test with more (abrupt) drifts using the following:

```python
np.random.seed(1)
mu, sigma = 0, 0.1  # mean and standard deviation
d_1 = np.random.normal(mu, sigma, 500) > 0
mu, sigma = 0.25, 0.1  # mean and standard deviation
d_2 = np.random.normal(mu, sigma, 500) > 0
mu, sigma = 0.5, 0.1  # mean and standard deviation
d_3 = np.random.normal(mu, sigma, 500) > 0
mu, sigma = 0.75, 0.1  # mean and standard deviation
d_4 = np.random.normal(mu, sigma, 500) > 0
data_stream = np.concatenate((d_1.astype(int), d_2.astype(int), d_3.astype(int), d_4.astype(int)))
```

With the following results from MOA:

|        |  MOA | scikit-multiflow |
|:------:|:----:|:----------------:|
| HDDM_A | 518,  1551 |       ?       |
| HDDM_W | 520, 1237, 1527 |       ?       |

Please use the above information to compare it with your implementation.

---
3. Finally, although coverage from your tests is very good., there are some lines of unreached code  (see [here](https://codecov.io/gh/scikit-multiflow/scikit-multiflow/pull/161/diff?src=pr&el=tree#diff-c3JjL3NrbXVsdGlmbG93L2RyaWZ0X2RldGVjdGlvbi9oZGRtX2EucHk=)) which can be easily covered. 

The easiest ones to reach are in the `reset` method of `HDDM_A`. Please add it to your test.

For the rest, maybe using a more complex stream can help. [Optional]

---

**Note:** You need to sync your PR with the master since some CI tests are failing due to changes in our dependencies. You will need the fix from the master in order for the unrelated CI tests to pass.

Let me know if I can be of help.

jiahy0825:
Thanks for your comments, and that helps me a lot. @jacobmontiel 

I have made some changes according to your comments:

1.  The documentation states that they return `self`:

   I have changed the documentation and delete inappropriate comments (such as comments below). Compared to add `return` at the end of the method, I think this change can make code more concise.

```
    Returns
    -------
    HDDM_W
    self
```

2. Tests between MOA and this python code. 

   First of all, I tested my implementation based on the drifts you suggested. But the result is not very well, my implementation can only find one drift at 515. 

   I conducted an analysis of the result:

   ```
   mu, sigma = 0.5, 0.1  # mean and standard deviation
   d_3 = np.random.normal(mu, sigma, 500) > 0
   ```

   According to your drifts: 
   
   For example, mean and standard deviation of d_3 is 0.5 and 0.1. It means that most data in d_3 is around 0.5, after compare d_3's data with 0, almost all data in d_3 changed to 1(`True`) . d_4 is the same as d_3. This phenomenon causes almost no conceptual drift in d_3 and d_4 (even in d_2). So I think my result of only detected one drift is reasonable.

   In order to do tests with more abrupt drifts, I used the following code, and did some tests between MOA and my implementation. 

   ```
   mu, sigma = 0.0, 0.1  # mean and standard deviation
   d_1 = np.random.normal(mu, sigma, 500) > 0
   mu, sigma = 0.25, 0.1  # mean and standard deviation
   d_2 = np.random.normal(mu, sigma, 500) > 0
   mu, sigma = 0.0, 0.1  # mean and standard deviation
   d_3 = np.random.normal(mu, sigma, 500) > 0
   mu, sigma = 0.25, 0.1  # mean and standard deviation
   d_4 = np.random.normal(mu, sigma, 500) > 0
   data_stream = np.concatenate((d_1.astype(int), d_2.astype(int), d_3.astype(int), d_4.astype(int)))
   ```

   Besides, I am curious about the problem of `delta=2`  between MOA and my code, I think The source of the problem is random  generators between Java and Python. So I generate drift stream first and write the stream to the file. Then MOA and scikit-multiflow read the same stream file and begin their test. (The data file is uploaded as an attachment)

   In the end, the results between MOA and scikit-multiflow are the same, which means `delta=0`

   Single drift:
[drift_stream_single.txt](https://github.com/scikit-multiflow/scikit-multiflow/files/3932763/drift_stream_single.txt)

|        | MOA  | scikit-multiflow |
| ------ | ---- | ---------------- |
| HDDM_A | 1052 | 1018             |
| HDDM_W | 1052 | 1018             |


   Abrupt drift:

[drift_stream_abrupt.txt](https://github.com/scikit-multiflow/scikit-multiflow/files/3932762/drift_stream_abrupt.txt)


|        | MOA       | scikit-multiflow |
| ------ | --------- | ---------------- |
| HDDM_A | 515, 1514 | 515, 1514        |
| HDDM_W | 518, 1513 | 518, 1513        |


3. I add `reset` to my test file and hope it helps.

If you have questions, please let me know. ​:slightly_smiling_face:​

jiahy0825:
After I push my new test code, I got `The Travis CI build failed`. (Python 3.7.4 on Windows failed)
But I only changed the test file, there is no reason to report an error.

I'm not very familiar with this problem, maybe I do something wrong when I sync my PR with the master? Can you help me or give me some suggestion?

Thanks a lot.

jacobmontiel:
Thanks for the changes and the tests @jiahy0825.

I have marked as resolved the remaining items from the review.

The last results are very helpful to validate your implementation, great work. Regarding the test stream with multiple drifts, it is strange because I also saved it into a file so both implementations would observe the same data. Perhaps I did something wrong. Do you have any comment on this aspect?

jiahy0825:
Thanks for your comments @jacobmontiel .
Regarding your question, I have the following comments:

1. Check the parameters and code that generated the data stream.
    For example, whether the mean and variance of the four streams d_1, d_2, d_3, and d_4 are correct. And whether the data is compared to 0 at the end.
2. Check the generated result, the generated data file should only contain 0 (`False`, data is negative) or 1 (`True`, data is positive).
3. I use python to generate the data and write to a file, maybe java's implementation is a little different.
4. Keep the MOA version up to date (although not necessarily useful).

The code that generates the data is as follows:

```python
# Data
np.random.seed(1)
mu, sigma = 0.0, 0.1  # mean and standard deviation
d_1 = np.random.normal(mu, sigma, 500) > 0
mu, sigma = 0.25, 0.1  # mean and standard deviation
d_2 = np.random.normal(mu, sigma, 500) > 0
mu, sigma = 0.5, 0.1  # mean and standard deviation
d_3 = np.random.normal(mu, sigma, 500) > 0
mu, sigma = 0.75, 0.1  # mean and standard deviation
d_4 = np.random.normal(mu, sigma, 500) > 0
data_stream = np.concatenate((d_1.astype(int), d_2.astype(int), d_3.astype(int), d_4.astype(int)))

# open file
fd = open('data_stream.txt', 'w')
# write data
for data in data_stream:
    fd.write(str(data) + "\n")
fd.close()
```

The corresponding data file is generated in this way:

[data_stream.txt](https://github.com/scikit-multiflow/scikit-multiflow/files/3939009/data_stream.txt)

By analyzing this data file, we can know that 1495 of the last 1500 data(d_2, d_3 and d_4) are 1, so intuitively, there should be only one concept drift, and it occurs between d_1 and d_2.

If you still have questions, maybe you can send the code and data files here and let us check together.

jacobmontiel:
Thanks for the detailed information @jiahy0825.

Looking into the steps you listed we are on the same page on the Python side. Perhaps the difference came when setting the test in MOA. I converted the data to an arff file and loaded it into MOA, perhaps there is an easier way. Comments?

Looking into the data, I agree that your test setup is clearer by repeating only two concepts.

jiahy0825:
Thanks for your comments @jacobmontiel.

I agree with you that perhaps the difference came when setting the test in MOA. I used a more direct approach.

Read the txt file directly, and then call the java class `HDDM_A_TEST`. The following is my code：

```java
public class HDDM_A_TEST_STREAM {
    public static void main(String[] args) {
        HDDM_A_Test hddm_a = new HDDM_A_Test();
        String str;
        ArrayList<Integer> arr = new ArrayList<Integer>();
        try {
            BufferedReader in = new BufferedReader(new FileReader("filepath\\data_stream.txt"));
            while ((str = in.readLine()) != null){
                arr.add(Integer.parseInt(str));
            }
        }catch (IOException e){
            System.out.println(e.getMessage());
        }
        for (int i = 0;i < arr.size();i++){
            hddm_a.input(arr.get(i));
            if (hddm_a.isChangeDetected) {
                System.out.println("change detected:" + i);
            }
        }
    }
}
```

I'm not very familiar with java, Hope it helps.

If there are any other questions, please feel free to point out.

jacobmontiel:
Thanks for the information, it helps a lot. Everything seems right. I will merge the PR now.

Thanks again @jiahy0825 for the hard work, your understanding and patience through the PR review.

Issue #133, opened 2019-08-28
[FEATURE] Time series forecasting methods


mertozer94:
I came across the need to implement forecasting methods at my work place. After doing some research, I realized that we even have this functionality in [excel](http://www.real-statistics.com/time-series-analysis/basic-time-series-forecasting/excel-2016-forecasting-functions/) and [libre office](https://wiki.documentfoundation.org/ReleaseNotes/5.2#New_spreadsheet_functions). My first instinct was to do a research on the existing algorithms. I have came upon the following ones; [Holt-Winters](https://en.wikipedia.org/wiki/Exponential_smoothing#Triple_exponential_smoothing) and [ARIMA](https://en.wikipedia.org/wiki/Autoregressive_integrated_moving_average).
 Both algorithms are good to forecast seasonal, trendy and nor seasonal nor trendy data. While naive Holt-Winter can [natively](https://github.com/influxdata/influxdb/pull/6621) be applied to streaming data, ARIMA needs to adapt to online learning, which can be found in [this paper](https://dl.acm.org/citation.cfm?id=3016160). 

I have looked upon some benchmarks and some comparisons online, and what I have seen was that both algorithms were used by developers. I believe we can benefit from having  (at least one) forecasting algorithm on time series data.  

Maybe it's my point of view, but I have found Holt-Winter easier to read and implement. What I have suggest is to implement Holf-Winter algorithm with having the following optional parameters, alpha, beta and gamma to forecast not seasonal nor trendy, trendy and seasonal data respectively. To be more clear, having only parameter alpha will create a model which will only apply Simple smoothing, having alpha and beta will apply Double smoothing, and having all together will apply Triple smoothing which is the classical Holf-Winter.
 
Also, I found this topic really interesting and I am willing to take the responsibility to develop this functionality.

jacobmontiel:
Thanks for the suggestion @mertozer94 

I am trying to figure out where/how we could fit this type of methods in the current structure.

Any comment @jmread?

jmread:
Time series forecasting is essentially just a multi-target classification/regression problem, where the targets are forecast values of the future of the same input. Perhaps it is possible to incorporate this into some generic wrapper/filter? transforming a raw stream of inputs x, into a data stream x, y1,y2,...,yk where k is defined as a parameter to the filter (number of steps to forecast). I believe this would allow sufficient flexibility to implement/adapt most methods in a data-stream context. Would something like that work? 
[...]

PR #162, opened 2019-11-23
Add a generic wrapper for forecasting classes.

mertozer94:
[Issue](https://github.com/scikit-multiflow/scikit-multiflow/issues/133)

This is a base class where raw stream of inputs x are transformed into a data stream x, y1,y2,...,yk where k is defined as a parameter to the number of steps to forecast and the training is done such as (x=x[t-1], y=x[t+k]) at moment t. Also training of the model is trivial to the user.

* A new base class for forecasting. 

---

Checklist

- [ ] Code complies with PEP-8 and is consistent with the framework.
- [ ] Files changed (update, add, delete) are in the PR's scope (no extra files are included).

### DIFF
--- a/src/skmultiflow/core/__init__.py
+++ b/src/skmultiflow/core/__init__.py
@@ -5,12 +5,15 @@ The :mod:`skmultiflow.core` module covers core elements of `scikit-multiflow`.
 from .base import BaseSKMObject
 from .base import ClassifierMixin
 from .base import RegressorMixin
+from .base import ForecasterMixin
 from .base import MetaEstimatorMixin
 from .base import MultiOutputMixin
 from .base import clone
 from .base import is_classifier
 from .base import is_regressor
+from .base import is_forecaster
 from .pipeline import Pipeline
 
-__all__ = ["BaseSKMObject", "ClassifierMixin", "RegressorMixin", "MetaEstimatorMixin",
-           "MultiOutputMixin", "Pipeline", "clone", "is_classifier", "is_regressor"]
+__all__ = ["BaseSKMObject", "ClassifierMixin", "RegressorMixin",
+           "ForecasterMixin", "MetaEstimatorMixin", "MultiOutputMixin",
+           "Pipeline", "clone", "is_classifier", "is_regressor", "is_forecaster"]
--- a/src/skmultiflow/core/base.py
+++ b/src/skmultiflow/core/base.py
@@ -636,6 +636,121 @@ class RegressorMixin(metaclass=ABCMeta):
                         multioutput='variance_weighted')
 
 
+class ForecasterMixin(metaclass=ABCMeta):
+    """Mixin class for all forecasting estimators in scikit-multiflow."""
+    _estimator_type = "forecaster"
+    last_samples = []
+    k = None
+    l = None
+
+    def __init__(self, k=1, l=1):
+        self.k = k
+        self.l = l
+
+    def partial_fit(self, X, y, sample_weight=None):
+        """ Partially (incrementally) fit the model with the following formula (x=x[t-k:t-1], y=x[t::t+l]).
+
+        Parameters
+        ----------
+        X : numpy.ndarray of shape (n_samples, n_features)
+            The features to train the model.
+
+        y: numpy.ndarray of shape (n_samples)
+            An array-like with the target values of all samples in X.
+
+        sample_weight: numpy.ndarray of shape (n_samples), optional (default=None)
+            Samples weight. If not provided, uniform weights are assumed. Usage varies depending on the learning method.
+
+        Returns
+        -------
+        self
+
+        """
+        if self.last_samples is None or len(self.last_samples) < self.k:
+            self._partial_fit(X, y, sample_weight=sample_weight)
+        else:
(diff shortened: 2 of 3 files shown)


jacobmontiel:
Thanks for the PR @mertozer94 

Pinging @jmread to get feedback on this.

**Note:** `codcov` error is due to missing tests for the new code.

jmread:
This looks promising!
In your comment above, I think you mean (x=x[t-1], y=x[t+k])? (To take into account k). Or you do mean (x=x[t-k:t-1], y=x[t]). Or a combination of these (there is one k or two?).

mertozer94:
> Thanks for the PR @mertozer94
> 
> Pinging @jmread to get feedback on this.
> 
> **Note:** `codcov` error is due to missing tests for the new code.

Thanks for the advice I will add some test cases as soon as possible. Also next time I will try to run locally 'codecov' but are there any other things that I should better check before pushing my commits? Like code coverage etc .

mertozer94:
> This looks promising!
> In your comment above, I think you mean (x=x[t-1], y=x[t+k])? (To take into account k). Or you do mean (x=x[t-k:t-1], y=x[t]). Or a combination of these (there is one k or two?).

Thank you for the response. You are right, I meant (x=x[t-1], y=x[t+k]). I will modify it right now. But I am open for discussion on this subject.

mertozer94:
> > This looks promising!
> > In your comment above, I think you mean (x=x[t-1], y=x[t+k])? (To take into account k). Or you do mean (x=x[t-k:t-1], y=x[t]). Or a combination of these (there is one k or two?).
> 
> Thank you for the response. You are right, I meant (x=x[t-1], y=x[t+k]). I will modify it right now. But I am open for discussion on this subject.

Actually, I think I have responded a bit quickly. What I had in my mind first was the training was done via (x=x[t-1], y=x[t]), and since I was thinking that forecasting was basically k times predicting the next value we could simply call predict function repeatedly k times. I now see that this is not possible.

 I realize that I had in mind (x=x[t-k:t-1], y=x[t]). As I said, I am open for discussion on this one.

jmread:
> I realize that I had in mind (x=x[t-k:t-1], y=x[t]). As I said, I am open for discussion on this one.

I think the most generic option is (x=x[t-k:t-1], y=x[t::t+l]) to allow multi-step forecasting into the future since multi-output learning is also an integral part of scikit-multiflow (of course, l=1 would be a nice default). But maybe the t:t+l part could be an extension of the wrapper.

mertozer94:
> > I realize that I had in mind (x=x[t-k:t-1], y=x[t]). As I said, I am open for discussion on this one.
> 
> I think the most generic option is (x=x[t-k:t-1], y=x[t::t+l]) to allow multi-step forecasting into the future since multi-output learning is also an integral part of scikit-multiflow (of course, l=1 would be a nice default). But maybe the t:t+l part could be an extension of the wrapper.

I agree that this is more generalized. But this means that, users needs to specify `k` and `l` before the training phase. Like you said, if they are not present, we can have some default values for them.

jmread:
> 
> I agree that this is more generalized. But this means that, users needs to specify `k` and `l` before the training phase. Like you said, if they are not present, we can have some default values for them.

Yes, I believe that `k=1` and `l=1` will make good defaults. 

There are numerous options that could be considered to extend this even further. For example, y=x[t+l] (a single, but long-range forecast). This has some relation to delayed labels. But this kind of flexibility could be added later. Even the simplest case with k=1,l=1 will open up many possibilities for the framework leveraging existing methods.

mertozer94:
Hey @jmread, I have included changes. Although, for the unit test, I can only imagine to test it via it's sub classes. Do you have any recommendation on that?

jmread:
> Hey @jmread, I have included changes. Although, for the unit test, I can only imagine to test it via it's sub classes. Do you have any recommendation on that?

Other than testing the handling of different values for the parameters k, l, indeed most of the testing can be done with the actual methods employed for the forecasting.

mertozer94:
> > Hey @jmread, I have included changes. Although, for the unit test, I can only imagine to test it via it's sub classes. Do you have any recommendation on that?
> 
> Other than testing the handling of different values for the parameters k, l, indeed most of the testing can be done with the actual methods employed for the forecasting.

I have added unit tests without creating a subclass.

mertozer94:
Hello @jacobmontiel @jmread It's been a long time, but do you have any idea on how we can improve the code coverage ?

> > > Hey @jmread, I have included changes. Although, for the unit test, I can only imagine to test it via it's sub classes. Do you have any recommendation on that?
> > 
> > 
> > Other than testing the handling of different values for the parameters k, l, indeed most of the testing can be done with the actual methods employed for the forecasting.
> 
> I have added unit tests without creating a subclass.

AnkurDebnath35:
@mertozer94 Can you share a sample code for the forecasting model which can handle stream data in an online fashion and is compatible with other functions of multiflow, like DriftDetectors?

AnkurDebnath35:
@mertozer94 I am unable to checkout this commit, throws out a "fatal:reference is not a tree" error while using the correct sha. Can someone help?

PR #169, opened 2019-12-02
[FEATURE] Bootstrap sampling option for the Hoeffding Adaptive Tree Classifier

jacobmontiel:

### Context

Performing bootstrap sampling in the leaf nodes improves (in most cases) the performance of the Hoeffding Adaptive Tree Classifier. This was enabled in version 0.4.0:

https://github.com/scikit-multiflow/scikit-multiflow/blob/cda70f059ca4cf1b6c9b7fe9bbcc70dcc5e480d4/src/skmultiflow/trees/hoeffding_adaptive_tree.py#L313-L315

This strategy is an improvement over the original method, consequently, it should be **optional** in our implementation in order for the user to decide whether to use it or not.

### Changes
This PR updates the signature of the Hoeffding Adaptive Tree Classifier to include a new `bootstrap_sampling` option:

```python
def __init__(self,
                 max_byte_size=33554432,
                 memory_estimate_period=1000000,
                 grace_period=200,
                 split_criterion='info_gain',
                 split_confidence=0.0000001,
                 tie_threshold=0.05,
                 binary_split=False,
                 stop_mem_management=False,
                 remove_poor_atts=False,
                 no_preprune=False,
                 leaf_prediction='nba',
                 nb_threshold=0,
                 nominal_attributes=None,
                 bootstrap_sampling=True)
````

The default value is set to `True` since, in general, it results in better classification performance.

---

Checklist

### DIFF
--- a/src/skmultiflow/trees/hoeffding_adaptive_tree.py
+++ b/src/skmultiflow/trees/hoeffding_adaptive_tree.py
@@ -65,10 +65,16 @@ class HAT(HoeffdingTree):
     nominal_attributes: list, optional
         List of Nominal attributes. If emtpy, then assume that all attributes are numerical.
 
+    bootstrap_sampling: bool, optional (default=True)
+        If True, perform bootstrap sampling in the leaf nodes.
+
     Notes
     -----
     The Hoeffding Adaptive Tree [1]_ uses ADWIN [2]_ to monitor performance of branches on the tree and to replace them
     with new branches when their accuracy decreases if the new branches are more accurate.
+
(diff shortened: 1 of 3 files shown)


smastelini:
Nice improvement @jacobmontiel. Everything seems fine to me.

Was not this modification applied to `RegressionHAT` in version `0.4.0`? Should this extension be a valid concern for the scope of this PR? We also could address this extension in the future.

Aside from that, I believe we can merge the PR. :D

jacobmontiel:
Good catch @smastelini. This improvement is not implemented in the Hoeffding Adaptive Tree Regressor. I will create a separate issue to discuss it. I have updated the PR to make clear that this only applies to the classifier version.

PR #172, opened 2019-12-03
[ENHANCEMENT] Making the `describe_subtree` method also consider the regression case

smastelini:
### Context

The `get_model_description` from the trees is used to print a representation of the current state of the model. This method, in turn, ends up using `describe_subtree` from `Node` (`skmultiflow.trees.nodes.py`). 

Although the current implementation has an option to print the statistics for regression models, this option is never reached. This fact occurs because the expected `ValueError` exception is not raised when accessing the node's statistics:

https://github.com/scikit-multiflow/scikit-multiflow/blob/9403e20150646371fb2dfff0187d3ddb05f25f3b/src/skmultiflow/trees/nodes/node.py#L153-L164

This PR fixes that problem by checking the `_estimator_type` property in the trees.

---

* Makes the Regression Hoeffding Trees properly printing their leaf statistics
* Add a new test in `RegressionHoeffdingTree` to cover the changes

---

Checklist


smastelini:
The failing in the automatic tests seem to be related with #173.

jacobmontiel:
Reported errors in Travis are related to the last version of scikit-learn (0.22). A temporary fix is available in the master branch. Please merge this fix with your branch.

smastelini:
The test failure seems to relate to numeric imprecision differences with different Python versions. All the tests had passed in my local testing environments.

To solve that issue, I added a `SequenceMatcher` call (from the `difflib` module) to calculate a percentage of agreement between the compared strings. I set an acceptable minimum amount of concordance to 90%. Nonetheless, it seems that for MacOS the threshold would around 88%.

Any comments on that strategy @jacobmontiel?

jacobmontiel:
I like the idea of using `difflib` since this kind of small differences are likely to keep showing.

I also suggest formatting the output for regression. This can help reduce the differences and is more readable:

```python
text = '{'
for i, (k, v) in enumerate(self._observed_class_distribution.items()):
    text += '{}: {:.4f}'.format(k, v)
    text += ', ' if i < len(self._observed_class_distribution) - 1 else ''
text += '}'
buffer += 'Statistics {}\n'.format(text)
```

Sorry for the hacky solution. There might be a more elegant way.

smastelini:
Agreed. I think that using four decimal places is a good compromise for visualization.

I will check this solution and also verify how it performs in the multi-target case.

Thanks for the suggestion @jacobmontiel.

PR #181, opened 2019-12-17
Use Numpy infix notation (& clean up code a bit).

AlbandeCrevoisier:

* Use Numpy infix notation
* Clean up surrounding code (could be broken in another PR)

---

Checklist


src/skmultiflow/trees/hoeffding_adaptive_tree_regressor.py
```diff
@@ -165,12 +165,12 @@ def _partial_fit(self, X, y, weight):
         self.sum_of_values += weight * y
         self.sum_of_squares += weight * y * y
 
-        try:
-            self.sum_of_attribute_values = np.add(self.sum_of_attribute_values, np.multiply(weight, X))
-            self.sum_of_attribute_squares = np.add(self.sum_of_attribute_squares, np.multiply(weight, np.power(X, 2)))
-        except ValueError:
-            self.sum_of_attribute_values = np.multiply(weight, X)
-            self.sum_of_attribute_squares = np.multiply(weight, np.power(X, 2))
+        if self.sum_of_attribute_values.size != 0:
```
jacobmontiel: Using `try` is a trick to improve performance when there are few chances of the error being raised. In this case, the `if` condition will only be evaluated as `False` once for initialization. Thus the usage of `try`.

tests/trees/test_regression_hoeffding_adaptive_tree.py
```diff
@@ -65,38 +63,48 @@ def test_hoeffding_tree_perceptron():
...
-                                       -64.23638301570358, 587.9771578214296, 649.8395655757931, 481.01214222804026,
-                                       305.4402728117724, 266.2096493865043, -445.11447171009775, -567.5748694154349,
-                                       -68.70070048021438, -446.79910655850153, -115.892348067663, -98.26862866231015,
-                                       71.04707905920286, -10.239274802165584, 18.748731569441812, 4.971217265129857,
-                                       172.2223575990573, -655.2864976783711, -129.69921313686626, -114.01187375876822,
-                                       -405.66166686550963, -215.1264381928009, -345.91020370426247, -80.49330468453074,
-                                       108.78958382083302, 134.95267043280126, -398.5273538477553, -157.1784910649728,
-                                       219.72541225645654, -100.91598162899217, 80.9768574308987, -296.8856956382453,
-                                       251.9332271253148])
+
```
jacobmontiel: Why are results different. This is also reflected in the MAE. Changing the operators should not affect the results.

src/skmultiflow/trees/hoeffding_adaptive_tree_regressor.py
AlbandeCrevoisier: Ah, ok. We used the same technique in embedded systems.
Should we add a comment to make it clear that we know when and why the `try` fails?

tests/trees/test_regression_hoeffding_adaptive_tree.py
AlbandeCrevoisier: Seeing the predictions, it looks like the computing precision is improved (differences are smaller than about 1e-05) so I assume it's either an earlier cast or one to a type with better precision.
But I don't know why that happens...

AlbandeCrevoisier: @jacobmontiel I've been thinking about this & I wondered wether it truly mattered: if it's just a matter of casting, could we not simply switch to the much more readable infix notation, regardless of the computation difference?

AlbandeCrevoisier: @jacobmontiel this was due to `+=`, probably because it's inplace.

src/skmultiflow/trees/hoeffding_adaptive_tree_regressor.py
smastelini: Hi @AlbandeCrevoisier, that's a great idea.

tests/trees/test_regression_hoeffding_adaptive_tree.py
smastelini: Did the results reverted back to their previous form since you changed the operators? That's unfortunate that the syntax sugar is not treated as a `ufunc`.

smastelini:
@AlbandeCrevoisier, please merge the last changes from `master` into your branch so that we proceed with the reviewing process. Since the creation of this PR, @jacobmontiel and I found some bugs in the Hoeffding Tree Regressors. Very much alike (and unfortunately) you will need to re-apply the infix notation to some of the affected code portions.

Thanks for the hard work so far. Let me know if you need any assistance.

AlbandeCrevoisier:
Stale PR, see #238

PR #182, opened 2019-12-18
Beautify Hoeffding Tree regressor.

AlbandeCrevoisier:

* PEP-8 / PEP-257
* Use iterators when possilbe: `for x in X:`
* `return` or `continue` right after its corresponding test (avoids unnecessary indentation & sometimes computation)
* use `np.isclose(var, 0, atol=1e-04)` to avoid dividing by numbers that are too close to 0.
  /!\ 1e-04 is arbitrary /!\
* rename `sd_squared` in the more explicit `var (because it's a variance...) & `sd` in `std` to follow the pythonic convention instead of the R one.
* Bugfix: return `np.zeros(get_dimension(X)[0])` instead of `0.0` in `predict`
* rename the test file according to the enforced convention.

Note: this is the state of the file in my arf_regression branch, I just extracted it in a seperate PR to keep things clean-ish, but this means it all comes in one commit & it also has the changes from the Numpy infix PR. I've put it here for reference, it might be wiser to break it down, I'm not sure how to proceed here.

---

Checklist


AlbandeCrevoisier:
Was only used for reference.

PR #183, opened 2019-12-18
[BUG] Fixing vote aggregation in HAT, typos in HAT/HATR, and improper usage of ADWIN in the adaptive trees

### LINKED ISSUE #180: [BUG] HAT not building alternate trees
**Describe the bug**
I updated to the new dev version and compared perfomance of the HAT classifier with the MOA implementation and noticed, that scikit-multiflows HAT is not building any alternate trees on my data that contains multiple concept drifts. Below I plotted the Tree depths of HAT and HT in MOA and scikit-multiflow, tested on the same dataset. 

HAT / HT Tree Depth in MOA:
UNSW_drift_weka3

HAT / HT Tree Depth in scikit-multiflow:
sk_HAT_HT_depth

The difference between HAT and HT is only caused by the bootstrap_sampling of HAT, if I disable it, the tree structure is identical between HAT and HT.

From what I observed so far, error.change in ada_split_node.py never evaluates to True:
https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/nodes/ada_split_node.py#L85-L88

I am not sure, if the root cause for the lack of alternate trees being built lies in HAT or ADWIN implementation.

**To Reproduce**
I compared the performance between MOA and skmultiflow using a custom dataset with artificial drifts created by sorting by a feature that is strongly correlated with the label. I am still working on a MWE that uses ConceptDriftStream, but I am not sure about the configurations I should use in order to detect situations where ADWIN should detect a concept change. 

Although I tested some configurations with the ConceptDriftStream, I never was able to create an example where the HAT trains at least one alternate tree.

I am not sure, if the Stream Generators in MOA and scikit-multiflow are comparable for this matter.
Any suggestions on this are welcome!

**Additional context**
Additionally, the docstring in adwin.py does not match the behavior in the ada_split_node.py, one of these mixes up the interepretation of 0 and 1. However, this is not the cause of the problem described above, since changing this does not cause alternate trees being built.

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/drift_detection/adwin.py#L180-L184

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/nodes/ada_split_node.py#L68

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/nodes/ada_split_node.py#L75-L78

smastelini:
This PR addresses Issue #180. Changes are applied to both HAT and HATR.

### Vote aggregation

In the `HoeffdingAdaptiveTreeClassifier` (HAT) algorithm, both its split nodes and leaf nodes have prediction capabilities. As a consequence, all of them carry statistics about the classes. HAT aggregates the statistics of all nodes from its root to the target leaf to calculate votes for an instance.

The current implementation uses the built-in `update` function from the `dict` class to aggregate the votes. This ends up discarding previous information and only keeping the most recent statistics. The mentioned problem lies within the following segment of HAT's code:

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/hoeffding_adaptive_tree.py#L187-L201

The problem happens at line 200, specifically.

This PR fixes this problem by using a helper function that sums the values of two dictionaries with the same keys. A special thanks to @jacobmontiel for providing the helper function.

--------

### Standardization of boolean variable checking

Additionally, the current HAT and HATR implementation do not follow the [PEP8  style guide](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html) for comparing the values of boolean variables.

The usage of `boolean_variable is True` or `boolean_variable is False` in multiple statements, such as

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/nodes/ada_split_node.py#L76

can lead to unexpected results. In the last example, the evaluation returns `1.0` when `0.0` would be expected, and vice-versa.

This PR standardize the HAT and HATR implementations to use the approach suggested by PEP8 to test boolean variables:

- `if boolean_variable`
- `if not boolean_variable`

--------

### Fixing typos in HAT/HATR code

Lastly, this PR fixes two typos in the current HAT and HATR implementations. The first one appears in the `number_leaves` method:

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/nodes/ada_split_node.py#L35-L41

The `_children` variable is a `dict`. Hence, the actual nodes should be accessed by the `_children.values()`.

Moreover, in the following line, the usage of `alternateTree` is incorrect (`_alternate_tree` is correct):

https://github.com/scikit-multiflow/scikit-multiflow/blob/f33717b0807747651a2292f0645c7b78efe55dbc/src/skmultiflow/trees/nodes/ada_split_node.py#L110

---------

* Fixes #180
* Adds the helper function `add_values_in_dict` to `skmultiflow.utils.utils.py`
* Updates HAT to use the helper function
* Adds documentation and a test for the `add_values_in_dict` function
* Corrects improper usage of boolean variable testing in HAT and HATR
* Corrects typos in the code

---

Checklist


smastelini:
For completeness, I performed a benchmark test comparing the classification Hoeffding Tree implementations for both MOA and skmultiflow. To this end, I generated an instance of the [RBFDrift generator](https://www.cs.waikato.ac.nz/~abifet/MOA/API/classmoa_1_1streams_1_1generators_1_1_random_r_b_f_generator_drift.html) in MOA (`100k` instances, `5` classes, `10` drifting centroids, and `0.1` of speed of change). All the variants used `leaf_prediction='nba'` (and the equivalent for MOA).

The results are presented in the following chart. I considered the accuracy and depth of the trees as evaluation measures.

skmultiflow_x_moa

The differences between the tree depths for HAT might be related to small implementation differences. Another source for the differences could be ADWIN. This possibility is going to be investigated in the future.

jacobmontiel:
Thanks @smastelini for looking into this. The plots show a good degree of agreement between implementations. There are minor differences that are acceptable considering that implementations are not exactly the same.

src/skmultiflow/utils/utils.py
```diff
@@ -179,3 +179,38 @@ def is_scalar_nan(x):
...
+    """ Adds two dictionaries, summing the values of elements with the same key.
+
+    This function iterates over the keys of dict_b and adds their corresponding
+    values to the elements in dict_a. If dict_b has a (key, value) pair that
+    does not belong to dict_a, this pair is added to the latter dictionary.
+
+    Parameters
+    ----------
+    dict_a: dictionary to update.
+    dict_b: dictionary containing the new observations.
```
jacobmontiel: Suggestion: 
```
    dict_b: dictionary whose values will be added to `dict_a`.
```

src/skmultiflow/utils/utils.py
```diff
@@ -179,3 +179,38 @@ def is_scalar_nan(x):
...
+    inplace: boolean (default: False)
+        Determines whether the changes will be implemented in-place (directly
```
jacobmontiel: Suggested rewording:
```
        If `True`, the addition is performed in-place and results are stored in `dict_a`.
        If `False`, `dict_a` is not changed and the results are returned in a new dictionary.
```

src/skmultiflow/utils/utils.py
```diff
@@ -179,3 +179,38 @@ def is_scalar_nan(x):
...
+        on dict_a, when set to True) or in a copy of dict_a (default behavior).
+
+    Returns
+    -------
+        dict
+            A dictionary containing the result of the merging operation.
```
jacobmontiel: Suggested rewording:
```
            A dictionary containing the result of the operation.
            Either a pointer to `dict_a` or a new dictionary depending on parameter `inplace`.
```

smastelini: Addressed in [117dc8b](https://github.com/scikit-multiflow/scikit-multiflow/pull/183/commits/117dc8bddb3faad066df992197abe7ff0f876397).

src/skmultiflow/utils/utils.py
smastelini: Addressed in [117dc8b](https://github.com/scikit-multiflow/scikit-multiflow/pull/183/commits/117dc8bddb3faad066df992197abe7ff0f876397).

src/skmultiflow/utils/utils.py
smastelini: Addressed in [117dc8b](https://github.com/scikit-multiflow/scikit-multiflow/pull/183/commits/117dc8bddb3faad066df992197abe7ff0f876397).

smastelini:
Thanks for the suggestions @jacobmontiel. I added a new commit to address them.

jacobmontiel:
Changes are correct. Will merge into the master. 

Thanks again @smastelini

PR #188, opened 2019-12-21
Fix bug in normalization calculation and perceptron leafs in HoeffdingTreeRegressor

AlbandeCrevoisier:

* Rename `sd_squared` in the more explicit `var`
* Rename `sd` in `std` (R convention -> NumPy convention)
* Protect normalization computation against division by zero
* Add a private method to calculate standard deviation incrementally
* Fix training implementation of perceptron leaf nodes
* Update calculation when using perceptron leafs in HTR

---

Checklist


src/skmultiflow/trees/hoeffding_tree_regressor.py
```diff
@@ -197,12 +197,12 @@ def normalize_sample(self, X):
...
-                          / self.samples_seen) / self.samples_seen
-            if self.samples_seen > 1 and sd_squared >= 0:
+            var = ((self.sum_of_attribute_squares[i] - self.sum_of_attribute_values[i] ** 2
+                          / self.samples_seen) / self.samples_seen)
+            if not np.isclose(var, 0, atol=1e-08):
                 mean = self.sum_of_attribute_values[i] / self.samples_seen
-                sd = np.sqrt(sd_squared)
-                normalized_sample.append((X[i] - mean) / (3 * sd))
+                std = np.sqrt(var)
+                normalized_sample.append((X[i] - mean) / (3 * std))
```
jacobmontiel: What about:
```python
            try: 
                normalized_sample.append((X[i] - mean) / (3 * std))
            except ZeroDivisionError:
                normalized_sample.append(0.0)
```

Considering the chances of the division by zero actually happening.

src/skmultiflow/trees/hoeffding_tree_regressor.py
```diff
@@ -197,12 +197,12 @@ def normalize_sample(self, X):
         """
         normalized_sample = []
         for i in range(len(X)):
-            sd_squared = (self.sum_of_attribute_squares[i] - self.sum_of_attribute_values[i] ** 2
-                          / self.samples_seen) / self.samples_seen
-            if self.samples_seen > 1 and sd_squared >= 0:
+            var = ((self.sum_of_attribute_squares[i] - self.sum_of_attribute_values[i] ** 2
```
jacobmontiel: `self.sum_of_attribute_values[i] * self.sum_of_attribute_values[i]` is faster than `self.sum_of_attribute_values[i] ** 2`

AlbandeCrevoisier: Did not know that, dully noted. Thanks for the info!

src/skmultiflow/trees/hoeffding_tree_regressor.py
AlbandeCrevoisier: Indeed, that's neat. I'll do that now!

AlbandeCrevoisier: So, because NaN is actually a valid NumPy value, only a RuntimeWarning will be raised... Same for taking the square root of a negative number.

Issue #190, opened 2020-01-02
Inconsistent use of sample_weight vs weight parameters [BUG]


benman1:
Estimators use as a parameter either **weight** or **sample_weight** in the partial_fit() method. This gives an inconsistent API across estimators. In sklearn, it is always **sample_weight**. I think this is the convention, scikit-multiflow should follow. I can fix this, if there are no objections.

Thanks!

jacobmontiel:
Hello @benman1 

Thanks for identifying this. Could you provide an example where this inconsistency is present? Indeed this should be the same (`sample_weight`) across all estimators.

benman1:
Hi, Jacob

Thanks for responding. I've just looked back at this and only found two instances were this was inconsistent (for some reason, I thought there were more). I've created a PR that fixes this: [PR](https://github.com/scikit-multiflow/scikit-multiflow/pull/202).

Thanks!
Ben.

PR #202, opened 2020-01-18
minor: rn parameter name from weight to sample_weight

benman1:
Refers to this issue:
**Inconsistent use of sample_weight vs weight parameters [BUG] #190**

This should be very minor: partial_fit and fit functions, if they have sample weights, this parameter should have a consistent name. I found (only) two instances where the parameter is called **weight** instead of **sample_weight** (for some reason, I thought there were more). I changed these to sample_weight (+documentation) and checked that it doesn't break anything. 

I am very impressed by this library, and all the work all of you have been putting in. Perhaps to flag: I didn't find any tests using the weights. 


jacobmontiel:
Thanks for addressing this issue @benman1.

benman1:
My pleasure. You are all doing a really good job.

Issue #177, opened 2019-12-11
[BUG]ConceptDriftStream generate bug.


zhanghaooo:
Hi, there is a bug in ConceptDriftStream.next_sample() when it use DataStream as stream and drift_stream

![image](https://user-images.githubusercontent.com/35730721/70598435-d1f3b980-1c25-11ea-87fc-b3665d5ab634.png)

this is my code:
```python 
from skmultiflow.data import ConceptDriftStream
from skmultiflow.data import HyperplaneGenerator
from skmultiflow.data import DataStream

stream = HyperplaneGenerator(n_features=2)
stream.prepare_for_use()
X, y = stream.next_sample(100)

stream = ConceptDriftStream(stream=DataStream(X, y, name="stream1"),
                            drift_stream=DataStream(X, y, name="stream2"))

stream.prepare_for_use()
X, y = stream.next_sample(100)
```

jacobmontiel:
Thanks for reporting this @zhanghaooo and providing the MWE.

I can confirm the bug. The root cause is in the initialization of the concept drift stream. In this case, the two input streams are not initialized and their metadata is not available.

A temporary workaround is:

```python
from skmultiflow.data import ConceptDriftStream
from skmultiflow.data import HyperplaneGenerator
from skmultiflow.data import DataStream

stream = HyperplaneGenerator(n_features=2)
stream.prepare_for_use()
X, y = stream.next_sample(100)

stream1=DataStream(X, y, name="stream1")
stream2=DataStream(X, y, name="stream2")
stream1.prepare_for_use()
stream2.prepare_for_use()

stream = ConceptDriftStream(stream=stream1,
                            drift_stream=stream2)

stream.prepare_for_use()
X, y = stream.next_sample(100)
```

We will submit a fix on the development branch soon.

jacobmontiel:
Addressed in #205 and the fix is in 03651318e8b039a80c5058c09d021f2c750f98b6

PR #205, opened 2020-02-05
Deprecate prepare_for_use method in Stream class.

### LINKED ISSUE #204: [DEPRECATION] Deprecate prepare_for_use method in the Stream class
The `prepare_for_use` method adds an additional step during a stream initialization without providing additional functionality. Additionally, it has led to errors such as in #177 

### Current implementation:
```
stream = SomeStream()   # initialization but not ready to use
stream.preapare_for_use()
```

### Proposed change:
```
stream = SomeStream()   # initialization and ready to use
```

The proposal is to turn the existing `prepare_to_use` method into a private method to be called during initialization. The public method will become a `stub` with no impact on the stream. The rationale for this is:

1. The impact on current implementation is minimal.
2. The change is hidden from the user.
3. We will keep the public method for at least 2 minor versions as a `stub` method that will raise a `FutureWarning`.
4. Existing code will only require removing the deprecated `prepare_for_use`.

jacobmontiel:

- Stream objects are ready to use after instantiation
- Streamline initialization of instances of the Stream class
- Streamlined stream restart methods
- Remove redundant code
- Update documentation to reflect this change
- Make some methods private in the stream generators.
- Add temporarily a FutureWarning to let the user know that prepare_for_use will be removed
- Update tests cases
- Fixes #204 and #177

**Note:** This is change does NOT affect the functionality of related methods, instances of the `Stream` class.

---

Checklist


src/skmultiflow/data/base_stream.py
```diff
@@ -236,7 +237,10 @@ def is_restartable(self):
 
     def restart(self):
         """  Restart the stream. """
-        self.prepare_for_use()
+        self.current_sample_x = None
+        self.current_sample_y = None
+        self.sample_idx = 0
+        self._prepare_for_use()
```
smastelini: Would not this call raise a `NotImplementedError`?

jacobmontiel: the `Stream` class is an abstract class so it should not be used as-is. Additionally, `_prepare_for_use` is an `@abstractmethod` so it must be implemented by all extensions of the `Stream` class.

smastelini: You are right. Sorry, I got confused. The change is correct!

smastelini:
The changes are correct! Thanks for looking into that @jacobmontiel.

PR #223, opened 2020-03-13
[BUG] Fixing case where not all classes are observed by LearnPPNSE's members and adding early stop to LearnPP

### LINKED ISSUE #166: [BUG] LearnNSE fails when not all the classes are observed by its batch base models
For some datasets, learn_nse.py fails. The following error shows for kdd dataset:

Traceback (most recent call last):
  File "learn_nse.py", line 62, in <module>
    preds.append(clf.predict(x)[0])
  File "/scratch/root/anaconda/anaconda3/lib/python3.7/site-packages/skmultiflow/meta/learn_nse.py", line 264, in predict
    votes = self.predict_proba(X)
  File "/scratch/root/anaconda/anaconda3/lib/python3.7/site-packages/skmultiflow/meta/learn_nse.py", line 248, in predict_proba
    return self.__vote_proba(X, len(self.ensemble), self.classes)
  File "/scratch/root/anaconda/anaconda3/lib/python3.7/site-packages/skmultiflow/meta/learn_nse.py", line 222, in __vote_proba
    votes += self.ensemble_weights[i] * y_predicts
ValueError: operands could not be broadcast together with shapes (1,23) (1,2) (1,23)

The snippet of the code is as follows:

```python
    import numpy as np
    from skmultiflow.meta import LearnNSE

        clf = LearnNSE(n_estimators=n_estimators, window_size=max_window_size)
        clf.partial_fit(data[0].reshape(1,data.shape[1]), [label[0]], classes=np.unique(label))

        preds = []
        for i in range(data.shape[0]):
            if (i+1)%int((len(label)/20)) == 0:
                print('%d%%' % int(np.round((i+1.)/len(label)*100, 0)))
            x = data[i].reshape(1, data.shape[1])
            y = np.array([int(label[i])])
            preds.append(clf.predict(x)[0])
            clf.partial_fit(x, y)

```
The kdd dataset has been downloaded from https://github.com/vlosing/driftDatasets, where labels have been mapped to integer values. The map of labels is as follows:
normal:0
buffer_overflow:1
loadmodule:2
perl:3
neptune:4
smurf:5
guess_passwd:6
pod:7
teardrop:8
portsweep:9
ipsweep:10
land:11
ftp_write:12
back:13
imap:14
satan:15
phf:16
nmap:17
multihop:18
warezmaster:19
warezclient:20
spy:21
rootkit:22

smastelini:
This PR addresses #166 and also avoid "division by zero" errors in `LearnPPNSEClassifier`. Moreover, it also adds an early stop mechanism to `LearnPP`. This action is necessary when all instances are already correctly classified before all the ensemble members are trained.

Thanks @Aashi-Jindal, for pointing out the bug in `LearnPPNSEClassifier`.

* Fixes #166
* Add early stop strategy to `LearnPP` to avoid `NA` normalization errors (which happens when all instances are correctly classified before all ensemble members are fit) 

---

Checklist


src/skmultiflow/meta/learn_nse.py
```diff
@@ -265,11 +270,33 @@ def __vote_proba(self, X, t, classes):
...
+                            )
 
             res.append(votes.reshape(len(classes)))
         return np.array(res)
 
+    def _fill_missing_probs(self, probs, obs_classes, all_classes):
+        proba_ordered = np.zeros(
+            (probs.shape[0], all_classes.size), dtype=np.float
+        )
+        sorter = np.argsort(all_classes)
```
jacobmontiel: What about renaming `sorter` as `sorterd_classes` for clarity

src/skmultiflow/meta/learn_pp.py
```diff
@@ -177,8 +177,8 @@ def partial_fit(self, X, y, classes=None, sample_weight=None):
         return self
 
     def __fit_batch(self, X, y):
-        ensemble = [copy.deepcopy(self.base_estimator) for _ in range(self.n_estimators)]
-        normalized_errors = [1.0 for _ in range(self.n_estimators)]
+        ensemble = []
```
jacobmontiel: What is the rationale behind this change?

src/skmultiflow/meta/learn_nse.py
```diff
@@ -265,11 +270,33 @@ def __vote_proba(self, X, t, classes):
...
-                    votes += self.ensemble_weights[i] * y_predicts
+                    try:
+                        votes += self.ensemble_weights[i] * y_predicts
+                    except ValueError:
+                        # sklearn learner
+                        if hasattr(h, 'classes_'):
+                            obs_classes = h.classes_
+                        else:  # skmultiflow learner
+                            obs_classes = h.classes
+                        votes += self.ensemble_weights[i] * \
```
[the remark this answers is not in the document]
smastelini: That is a valid concern. Do you have an idea of how to implement that? In other words, how to retrieve the classes observed by an arbitrary classifier?

src/skmultiflow/meta/learn_pp.py
smastelini: Originally, the ensemble algorithm is designed to train all of its members and combine their predictions considering the observed normalized errors. However, when training the classifiers, if the observed prediction error shrinks to zero before all learners are fit, the error normalization results in a `ValueError: probabilities contain NaN` message. This happens because the error normalization factor is zero, so the resulting division (for weight normalization) is `NaN`. I created a [new test](https://github.com/scikit-multiflow/scikit-multiflow/blob/18b8f518f067e4739d850ddbd91e30fe5e7c3adb/tests/meta/test_learn_pp.py#L58-L100) to encompass this situation.

In the modified version, ensemble members are incrementally added until the maximum number of learner is reached or the observed error is zero.

src/skmultiflow/meta/learn_nse.py
smastelini: Totally agreed! I renamed the variable in commit [3ec9def](https://github.com/scikit-multiflow/scikit-multiflow/pull/223/commits/3ec9defdc5593530fffc588be8054e6c091805a0). Thanks for the suggestion @jacobmontiel.

src/skmultiflow/meta/learn_nse.py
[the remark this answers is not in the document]
jacobmontiel: I am thinking on two options:

1. Raise a warning :-)
2. Make the `classes` parameter in `partial_fit` a required parameter

src/skmultiflow/meta/learn_pp.py
jacobmontiel: This is a valid concern and a valid implementation-wise solution. However, my concern is that this is not part of the original algorithm. We could add a note in the code (basically what you wrote above). I would assume this only happens in rare occasions.

src/skmultiflow/meta/learn_nse.py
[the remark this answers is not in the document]
smastelini: I agree with both options. Particularly regarding option 2, I believe skmultiflow already uses this approach for some other algorithms. We could easily extend that to this case to keep the same standards. What do you think?

src/skmultiflow/meta/learn_pp.py
smastelini: Good idea @jacobmontiel and a valid observation. I added a note to `LearnPPClassifier` in commit [c460875](https://github.com/scikit-multiflow/scikit-multiflow/pull/223/commits/c460875b27a8aef9a70a33c14e94d82b54f8c045). In the note, I state the reason for the early stop mechanism and also mention that this situation is very specific. Thank you for pointing that out.

src/skmultiflow/meta/learn_nse.py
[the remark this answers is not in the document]
jacobmontiel: I prefer the first option. Given that this is a corner case and should be addressed by the user. On the other hand, making `classes` a required parameter has the downside of limiting the application to "emerging classes".

src/skmultiflow/meta/learn_pp.py
```diff
@@ -61,6 +61,19 @@ class LearnPPClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorMixin):
         'classes' parameter is not passed in the first partial_fit call, or if they
         are passed in further calls but differ from the initial classes.
 
+    Notes
+    -----
+        Originally, this ensemble algorithm was designed to train all of its members and combine
```
jacobmontiel: After re-reading this in the context of the docstring I suggest we avoid mentioning code details and focus on the algorithmic aspects. Suggested rewording:

```python
    """
    Notes
    -----
    Originally, Learn++ is designed to train all of its members and combine
    their predictions considering the observed normalized errors. However,
    when training the base estimators, if the observed prediction error
    shrinks to zero before all estimators are trained, the error normalization
    is ill-defined, i.e., the instance error-based weight normalization factor
    (the sum of the errors) is zero.
    This implementation adds an 'early stop' mechanism to circumvent this
    corner case: LearnPPClassifier stops adding members to the ensemble if
    all instances are correctly classified. Otherwise (its normal behavior),
    the model uses as many ensemble members as defined via the ``n_estimators``
    parameter.
    """
```

src/skmultiflow/meta/learn_nse.py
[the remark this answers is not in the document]
smastelini: I see your point. So, the idea is to check whether the learner belongs to either sklearn or skmultiflow, and in case none of them check, we raise a Warning. Do you have any specific kind of warning type in mind?

src/skmultiflow/meta/learn_pp.py
smastelini: Thank you for the suggestion @jacobmontiel. It is more concise and avoid too many implemention details. I updated the description in commit [f37bb96](https://github.com/scikit-multiflow/scikit-multiflow/pull/223/commits/f37bb967d5f08874b51eab805ab8a7fff4d56a28).

src/skmultiflow/meta/learn_nse.py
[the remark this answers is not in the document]
smastelini: Another option is to raise an error in case the learner does not comply with either `sklearn` or `skmultiflow` standards. Thanks for suggesting that, @jacobmontiel. This last alternative is available in commit [97fe7dc](https://github.com/scikit-multiflow/scikit-multiflow/pull/223/commits/97fe7dc3c1b62c95018ca7f0bbd597b7360b34c8).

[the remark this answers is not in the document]
jacobmontiel: Seems good to me. Thanks!

PR #226, opened 2020-03-26
Fix bug in LeverageBaggingClassifier that resulted in unreachable code during initialisation when using output codes

### LINKED ISSUE #224: This part will never be executed
https://github.com/scikit-multiflow/scikit-multiflow/blob/3e4bc6856775707c1b8e9e664d197887307e572f/src/skmultiflow/meta/leverage_bagging.py#L204

jacobmontiel:
- Enable unreachable code related to the matrix codes initialization
- Fix bug when enable_code_matrix options is enabled
- Rename algorithm to `LeveragingBaggingClassifier` for consistency with the paper
- Update tests and improve coverage by introducing tests for variants of the method
- Improve documentation
- Add two private methods to simplify code
- Updated code example
- Enforce PEP-8 character limit
- Fixes #224

---

Checklist

### DIFF
--- a/src/skmultiflow/meta/leverage_bagging.py
+++ b/src/skmultiflow/meta/leverage_bagging.py
@@ -63,8 +63,8 @@ class LeveragingBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorM
     
     Raises
     ------
-    ValueError: A ValueError is raised if the 'classes' parameter is not \
-    passed in the first ``partial_fit call``.
+    ValueError: A ValueError is raised if the ``classes`` parameter is not \
+    passed in the first ``partial_fit`` call.
 
     Notes
     -----
@@ -129,11 +129,11 @@ class LeveragingBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorM
     
     """
 
-    LEVERAGE_ALGORITHMS = ['leveraging_bag',
-                           'leveraging_bag_me',
-                           'leveraging_bag_half',
-                           'leveraging_bag_wt',
-                           'leveraging_subag']
+    _LEVERAGE_ALGORITHMS = ['leveraging_bag',
+                            'leveraging_bag_me',
+                            'leveraging_bag_half',
+                            'leveraging_bag_wt',
+                            'leveraging_subag']
 
     def __init__(self,
                  base_estimator=KNNClassifier(),
@@ -158,8 +158,10 @@ class LeveragingBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorM
         self.enable_code_matrix = enable_code_matrix
         self.w = w
         self.delta = delta
-        if leverage_algorithm not in self.LEVERAGE_ALGORITHMS:
-            raise ValueError("Leverage algorithm not supported.")
+        if leverage_algorithm not in self._LEVERAGE_ALGORITHMS:
+            raise ValueError("Invalid option for leverage_algorithm: '{}'\n"
+                             "Valid options are: {}".format(leverage_algorithm,
+                                                            self._LEVERAGE_ALGORITHMS))
         self.leverage_algorithm = leverage_algorithm
         self.random_state = random_state
         self.__configure()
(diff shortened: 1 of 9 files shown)


src/skmultiflow/meta/leverage_bagging.py
```diff
@@ -9,52 +9,62 @@
...
-        If None, the random number generator is the RandomState instance used by `np.random`.
+        If None, the random number generator is the RandomState instance used
+        by `np.random`.
     
     Raises
     ------
-    ValueError: A ValueError is raised if the 'classes' parameter is
-    not passed in the first partial_fit call.
+    ValueError: A ValueError is raised if the 'classes' parameter is not \
+    passed in the first ``partial_fit call``.
```
smastelini: Should not the markdown end after `partial_fit`?

src/skmultiflow/meta/leverage_bagging.py
```diff
@@ -265,16 +259,20 @@ def __partial_fit(self, X, y):
                 k = self._random_state.poisson(1)
```
smastelini: Sorry if the comment is misplaced (github doesn't let me comment on collapsed portions of the diff view). I think it would be nice to add short comments for each of the `if/elif/else` statements evaluating the values in `self.LEVERAGE_ALGORITHMS`. In this sense, we can easily track which Leveraging Bagging variant each portion of the code covers.

src/skmultiflow/meta/leverage_bagging.py
```diff
@@ -81,46 +91,48 @@ class LeverageBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorMix
...
-    >>> 
-    >>> # Displaying the results
-    >>> print(str(sample_count) + ' samples analyzed.')
+    ... # Displaying the results
+    ... print(str(sample_count) + ' samples analyzed.')
     2000 samples analyzed.
-    >>> print('LeverageBaggingClassifier performance: ' + str(corrects / sample_count))
-    LeverageBagging classifier performance: 0.945
+    >>> print('LeveragingBaggingClassifier performance: ' + str(corrects / sample_count))
+    LeveragingBagging classifier performance: 0.843
```
smastelini: Did the expected performance for this example change because of the modification on `SEAGenerator`'s setup?

src/skmultiflow/meta/leverage_bagging.py
```diff
@@ -81,46 +91,48 @@ class LeverageBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorMix
...
     
     """
 
-    LEVERAGE_ALGORITHMS = ['leveraging_bag', 'leveraging_bag_me', 'leveraging_bag_half', 'leveraging_bag_wt',
+    LEVERAGE_ALGORITHMS = ['leveraging_bag',
```
smastelini: Should this constant be private?

src/skmultiflow/meta/leverage_bagging.py
jacobmontiel: Looking into the file's history, it seems that the value corresponded to an older version without the `random_state` option.

jacobmontiel: Additionally, the SEA generator now takes values in [0-1] range thus we needed to update the example.

src/skmultiflow/meta/leverage_bagging.py
jacobmontiel: Fixed in e979d59d7b7aee316c049a64d4540c07114848bc

src/skmultiflow/meta/leverage_bagging.py
jacobmontiel: Nice catch. Added comments in e979d59d7b7aee316c049a64d4540c07114848bc

src/skmultiflow/meta/leverage_bagging.py
jacobmontiel: Yes, updated in e979d59d7b7aee316c049a64d4540c07114848bc

src/skmultiflow/meta/leverage_bagging.py
smastelini: Got it! Marking this question as solved. Thank you for checking it out.

PR #229, opened 2020-04-06
Minor updates for HalfSpaceTrees

jacobmontiel:
The current implementation of the `HalfSpaceTrees` includes the number of features (`n_features`) as a required parameter. This goes against the strategy followed in other implementations where such information is obtained from the data itself.

- Remove `n_features` as required parameter for initializing a `HalfSpaceTrees` instance. `n_features` is now obtained from the data as it is passed for training.
- Enforce PEP8 char line limit
- Minor cosmetic changes

---

#### Implementation

#### Tests

### DIFF
--- a/src/skmultiflow/anomaly_detection/half_space_trees.py
+++ b/src/skmultiflow/anomaly_detection/half_space_trees.py
@@ -198,7 +198,7 @@ class HalfSpaceTrees(BaseSKMObject, ClassifierMixin):
                 predictions.append(0)
             else:
                 # if prediction of this instance is greater than the threshold
-                # defined,then this instance is classified as an anomaly.
+                # defined, then this instance is classified as an anomaly.
                 if y_proba[0][1] > self.anomaly_threshold:
                     predictions.append(1)
                 else:
@@ -386,7 +386,7 @@ class HalfSpaceTree:
         dict (class_value, weight)
 
         """
-        if self.root is not None and self.is_learning_phase_on is not True:
+        if self.root is not None and not self.is_learning_phase_on:
             score = self.anomaly_score(X)
             anomaly_score = 1 - score / max_score
             return {0: 1 - anomaly_score, 1: anomaly_score}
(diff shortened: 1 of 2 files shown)


src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -516,11 +521,11 @@ def update_model(self, node):
...
         if node is not None:
-            if node.r is not 0 or node.l is not 0:
-                node.r = node.l
+            if node.right_mass is not 0 or node.left_mass is not 0:
+                node.right_mass = node.left_mass
 
-            if node.l is not 0:
-                node.l = 0
+            if node.left_mass is not 0:
+                node.left_mass = 0
```
smastelini, line 528: When I run my the tests locally, pytest warns me that we should use something like: node.rigth_mass != 0 rather than relying on is not 0. What do you think? Could we leverage this PR to apply this fix?

src/skmultiflow/anomaly_detection/half_space_trees.py
```diff
@@ -197,18 +197,19 @@ def predict(self, X):
                 # Ensemble is empty, all classes equal, default to zero
                 predictions.append(0)
             else:
-                # if prediction of this instance being anomaly is greater than the threshold defined,
-                # then this instance is classified as an anomaly.
+                # if prediction of this instance is greater than the threshold
+                # defined,then this instance is classified as an anomaly.
```
smastelini, line 201: We could add a space after the comma (pretty much an optional, cosmetic, change). What do you think?

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: Nice catch. I changed it in cc37f898c0cb91ed786c23f9413fad11b9929592 as well as other non-standard uses of `is not`.

src/skmultiflow/anomaly_detection/half_space_trees.py
jacobmontiel: Definetly. Updated in cc37f898c0cb91ed786c23f9413fad11b9929592

src/skmultiflow/anomaly_detection/half_space_trees.py
smastelini: Nice!

PR #236, opened 2020-04-21
Make EvaluatePrequential and EvaluateHoldout report all processed samples in the output file

smastelini:
This PR updates the default behavior of `EvaluatePrequential` when reporting the last measurements and writing the output file. The current implementation updates metrics until the last multiple of `n_wait` which might be less than the user-defined `max_samples`, provided that there are enough samples in the stream. For example, if `max_samples=2500`and `n_wait=200` then the last metrics update is performed at `sample_count=2400` 

#### Expected behavior

Evaluators (prequential and hold out) **must** include measurements for up to `max_samples` regardless if `n_wait` has been reached at the end of the evaluation. In the example above, a final metrics update must be performed for the remaining `100` samples in the stream, to get the final metrics at `sample_count=2500`.

---

#### Implementation

#### Tests

### DIFF
--- a/src/skmultiflow/evaluation/evaluate_holdout.py
+++ b/src/skmultiflow/evaluation/evaluate_holdout.py
@@ -285,7 +285,7 @@ class EvaluateHoldout(StreamEvaluator):
                     else:
                         perform_test = (self.global_sample_count - self.test_size) % self.n_wait == 0
 
-                    if perform_test | (self.global_sample_count >= self.max_samples):
+                    if perform_test | (self.global_sample_count >= actual_max_samples):
 
                         if self.dynamic_test_set:
                             print('Separating {} holdout samples.'.format(self.test_size))
--- a/src/skmultiflow/evaluation/evaluate_holdout.py
+++ b/src/skmultiflow/evaluation/evaluate_holdout.py
@@ -71,7 +71,7 @@ class EvaluateHoldout(StreamEvaluator):
         | 'average_mean_squared_error'
         | 'average_mean_absolute_error'
         | 'average_root_mean_square_error'
-        | **Experimental** (no plot generated)
+        | **General purpose** (no plot generated)
         | 'running_time'
         | 'model_size'
 
--- a/src/skmultiflow/evaluation/evaluate_holdout.py
+++ b/src/skmultiflow/evaluation/evaluate_holdout.py
@@ -243,7 +243,7 @@ class EvaluateHoldout(StreamEvaluator):
 
         performance_sampling_cnt = 0
         print('Evaluating...')
-        while ((self.global_sample_count < self.max_samples) & (self._end_time - self._start_time < self.max_time)
+        while ((self.global_sample_count < actual_max_samples) & (self._end_time - self._start_time < self.max_time)
                & (self.stream.has_more_samples())):
             try:
                 X, y = self.stream.next_sample(self.batch_size)
--- a/src/skmultiflow/evaluation/evaluate_prequential.py
+++ b/src/skmultiflow/evaluation/evaluate_prequential.py
@@ -355,7 +355,7 @@ class EvaluatePrequential(StreamEvaluator):
                             self.running_time_measurements[i].update_time_measurements(self.batch_size)
 
                     if ((self.global_sample_count % self.n_wait) == 0 or
-                            (self.global_sample_count >= self.max_samples) or
+                            (self.global_sample_count >= actual_max_samples) or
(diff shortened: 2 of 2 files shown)


jacobmontiel:
Thanks for the change @smastelini 
The same change applies for `EvaluateHoldout`, can you look into this?

smastelini:
>  The same change applies for EvaluateHoldout, can you look into this?

Sure! I'll check `EvaluateHoldout` and update the PR's description.

fabriciojoc:
Does It also impact EvaluatePrequentialDelayed in my PR (pull request [#222](https://github.com/scikit-multiflow/scikit-multiflow/pull/222))?

smastelini:
> Does It also impact EvaluatePrequentialDelayed in my PR (pull request #222)?

Hi @fabriciojoc, thanks for pointing that out. I believe that this might impact the delayed evaluator. At least the [`_update_metrics_delayed`](https://github.com/scikit-multiflow/scikit-multiflow/blob/23e9330ed7f9f559ca48ef6fa5650773aec06be0/src/skmultiflow/evaluation/evaluate_prequential_delayed.py#L297-L299) method should implement similar changes.

fabriciojoc:
> > Does It also impact EvaluatePrequentialDelayed in my PR (pull request #222)?
> 
> Hi @fabriciojoc, thanks for pointing that out. I believe that this might impact the delayed evaluator. At least the [`_update_metrics_delayed`](https://github.com/scikit-multiflow/scikit-multiflow/blob/23e9330ed7f9f559ca48ef6fa5650773aec06be0/src/skmultiflow/evaluation/evaluate_prequential_delayed.py#L297-L299) method should implement similar changes.

Thanks @smastelini! I will consider it when updating my PR, I still have some pending changes to do.

fabriciojoc:
By the way, do you think that it is valid to use [tdqm](https://tqdm.github.io/)? I discovered this lib this week and maybe this can be useful for the loading bar when processing samples. Sorry for posting this here, but maybe this PR has to do with it (or future work on it).

jacobmontiel:
> By the way, do you think that it is valid to use tdqm? 

Currently, we have a less sophisticated progress bar. It does the job and no dependencies are required. However, this alternative might be a nice improvement. Please create a new issue to asses how many people would be interested in this feature.

jacobmontiel:
> I could not find another place to update the script regarding the metrics' report. @jacobmontiel, could you please check whether am I forgetting something?

This is the line in charge of triggering the test:

https://github.com/scikit-multiflow/scikit-multiflow/blob/e93866ceaa765e36137217ccd7d27af1ab510732/src/skmultiflow/evaluation/evaluate_holdout.py#L288

Looking back at the change, I am now confused about why it works. This is where `actual_max_samples` is calculated:

https://github.com/scikit-multiflow/scikit-multiflow/blob/e93866ceaa765e36137217ccd7d27af1ab510732/src/skmultiflow/evaluation/evaluate_prequential.py#L277-L279

The original intention of this part of the code was to protect for cases where the stream might contain fewer samples than `self.max_samples`, which is not the case at hand.

smastelini:
I am now a little bit confused too. :P

From what I can gather, if the stream is infinite or has more available instances than the passed `max_samples`, the evaluators will limit the auxiliary variable `actual_max_samples` to the desired limit of samples to process.

The alternative case would be when `self.stream.n_remaining_samples() <= self.max_samples.` In this situation, the evaluator should process all the available instances. This is actually the scenario which this PR tackles. 

Are we missing another corner case?

> This is the line in charge of triggering the test:

> https://github.com/scikit-multiflow/scikit-multiflow/blob/e93866ceaa765e36137217ccd7d27af1ab510732/src/skmultiflow/evaluation/evaluate_holdout.py#L288

Nice catch! I missed that. Thank you for pointing that out.

If there is not a case we are missing, the following should be the solution:

``` python
           if perform_test | (self.global_sample_count >= actual_max_samples):
```

jacobmontiel:
> Are we missing another corner case?

I don't think so. 

The csv file is (actually) updated in two cases:

1. When the buffer size is about to be exceeded
https://github.com/scikit-multiflow/scikit-multiflow/blob/e93866ceaa765e36137217ccd7d27af1ab510732/src/skmultiflow/evaluation/base_evaluator.py#L573-L578

2. At the end of the evaluation, if and only if there are remaining lines in the buffer
https://github.com/scikit-multiflow/scikit-multiflow/blob/e93866ceaa765e36137217ccd7d27af1ab510732/src/skmultiflow/evaluation/evaluate_prequential.py#L371-L372

Am I missing something?

Issue #111, opened 2019-05-08
[BUG] VFDR test fails in Python 3.6


jacobmontiel:
**Describe the bug**
The VFDR test fails in Travis for Python 3.6 during the call to `get_measurements`

**To Reproduce**
https://github.com/scikit-multiflow/scikit-multiflow/blob/e455cb9023bda8c2a5a4ec7917c9b3c7ab8586a9/tests/rules/test_vfdr.py#L60-L65

**Expected behavior**
The test should pass on all supported versions.

**Additional context**
This bug was found during the review of #108 
The test was successfully run locally for Python 3.5, 3.6 and 3.7

remiadon:
@jacobmontiel should this be closed ? The PR seems to be already merged

jacobmontiel:
There seem to be some differences in the calculation of size objects for different OSes. This has been observed after upgrading the Travis configuration for macOS. Commit 74c1bed3dc4c3af15c97e723ab57708de8ee4434 includes a workaround in `test_calculate_object_size` to only asses object sizes on Linux systems. A similar approach should be applied here.

PR #108, opened 2019-04-22
New classifiers, evaluation metrics and fixes

Darkmyter:
Changes proposed in this pull request:

* New classifier: Very Fast Decision Rules ([paper](https://link.springer.com/article/10.1007/s10618-013-0340-z)).
* New meta classifiers: Online UnderOverBagging, Online Boosting, Online SMOTEBAgging, Online RUSBoost, Online Adac2, Online CSB2 ([paper](https://ieeexplore.ieee.org/document/7567497)).
* New split criterion: Hellinger Distance ([paper](https://arxiv.org/pdf/1405.2278.pdf)).
* New evaluation metrics: F1 score, precision, recall and g-mean. Only calculated for the positive class and in case of binary classification.
* Added two base classes: Rule and Predicate that form the rules of VFDR and possibly other algorithms such as: RILL and eRules.
* Added method for printing Hoeffding tree as list of rules or getting them.
* Corrected documentation for data generators.
* Fixed bug in LeverageBagging: set(self.classes) in place of self.classes. 
* Fixed bug in OzaBaging:  resampling parameter should be calculated after each example and not batch of examples.

---

Checklist

- [x] Code complies with PEP-8 and is consistent with the framework.
- [x] Code is properly documented.
- [x] Tests are included for new functionality or updated accordingly.
- [x] Travis CI build passes with no errors.
- [x] Test Coverage is maintained (threshold is -0.2%).
- [ ] Files changed (update, add, delete) are in the PR's scope (no extra files are included).

PR #237, opened 2020-04-22
[FEATURE] Add AdaptiveRandomForestRegressor and streamline attribute observer for regression

smastelini:
This PR resumes the work (PR #145) started by Alban to implement the `AdaptiveRandomForestRegressor` (ARFReg). The implementation was finalized and tested. I have also added some additional options, such as different aggregation and weighting strategies, and different kinds of setups to track concept drifts.

As base learners, the Forest uses trees that derive from `HoeffdingTreeRegressor` (HTR), in detriment of FIMT-DD (as in MOA). HTRs do not have inner concept drift adaptation mechanisms; however, such action is performed at higher abstraction levels, by the ensemble algorithm. This action repeats in the classification version of Adaptive Random Forests.

In addition to ARFReg, this PR also streamlines the usage of attribute observers for regression. It adds a new and improved observer (derived from the multi-target regression observer) that works with both single-target and multi-target tasks. This enhanced `NumericAttributeRegressionObserver` also enables the support to instance weighting, an essential aspect for bagging-based algorithms. Hence, just a single observer is used for all the regression trees.

* Finishes work initiated in PR #145
* Streamline regression observers

---

#### Implementation

#### Tests


jacobmontiel:
`test_hoeffding_tree_regressor_coverage` fails in Python 3.5. This seems related to the object size calculation. This is a known issue (#108 #111). We can deactivate the test for Python 3.5 or add some delta to account for the small difference.

smastelini:
> test_hoeffding_tree_regressor_coverage fails in Python 3.5. This seems related to the object size calculation. This is a known issue (#108 #111). We can deactivate the test for Python 3.5 or add some delta to account for the small difference.

I vote to add a delta. What do you think?

jacobmontiel:
Agree, a delta seems like a good compromise.

smastelini:
> Agree, a delta seems like a good compromise.

I added a small tolerance in the failing test. It's working now :)

smastelini:
The lack of coverage relates mainly to this portion:

https://github.com/smastelini/scikit-multiflow/blob/f0b060d1e6874f8f5b402a7c142693591c5f1460/src/skmultiflow/trees/attribute_observer/numeric_attribute_regression_observer.py#L160-L231

I have ported this helper function from the previous attribute observer. Nonetheless, none of the trees actually use it. This method removes potentially bad split candidates from the binary search tree using a bottom-up strategy.

smastelini:
## Benchmark tests

I also performed some benchmark tests with the new implementation. Some tests are still running, and I'll update this message as they complete. I mainly used default parameters during the tests for both MOA and skmultiflow. I also set up MOA to use the same `delta` value applied in skmultiflow for the ADWIN objects (drift and warning detectors). In the following charts, I use the following notation:

* subscript `m`: indicates that the trees are using the target mean to provide responses at leaves.
* subscript `p`: the trees are using linear perceptrons to provide responses.
* the `w` prefix indicates that individual trees' responses are weighted (I setup the forests to use `mae` as the weighting scheme -- please refer to the documentation).
* unless otherwise stated (with the `+median` suffix), all forests are aggregating individual trees' predictions with the mean operation.
* the forests are monitoring the Mean Absolute Error to track concept drifts.

---

### Fried dataset

I started with the well-known synthetic task `Fried`. Here I compare multiple variants of ARFReg against the original MOA's version:

arfr_error_comparison_fried

The performance of `leaf_prediction=perceptron` and `aggregation_method='median'` led me to set this combination as the default. Thank you for this suggestion, @hmgomes.

---

### Hyper datasets

Next, I chose two variants of ARFReg and applied them to versions of the `Hyper` dataset (another well-known synthetic task) containing different kinds of concept drift. The selected versions were:

1. `leaf_prediction='mean'`, `aggregation_method='mean'`, and `weighted_vote_strategy=None`: this variant mimics the MOA implementation with its default parameters. The only major difference lies in the base learner choice (FIMT-DD in MOA and HoeffdingTreeRegressor in skmultiflow).
2. `leaf_prediction='perceptron'`, `aggregation_method='median'`, and `weighted_vote_strategy=None`: the best variant in the previous test.

#### Hyper dataset with Abrupt drifts

arfr_error_comparison_hyper_abrupt

#### Hyper dataset with Gradual drifts

arfr_error_comparison_hyper_gradual

jacobmontiel:
> I have adapted this helper function from the previous attribute observer. Nonetheless, none of the trees actually use it.

Is this relevant for the regression case?

smastelini:
> Is this relevant for the regression case?

This action is described in the original FIMT-DD's paper. I have checked the FIMT-DD code in MOA. The equivalent of the `remove_bad_splits` is called in case a split attempt fails:

https://github.com/Waikato/moa/blob/8b2f60d06958dfb1d2a6e1f94defb67e803b8971/moa/src/main/java/moa/classifiers/trees/FIMTDD.java#L899-L917

Nonetheless, FIMT-DD does not implement the strategy to completely disabling features as in`HoeffdingTreeRegressor`:

https://github.com/scikit-multiflow/scikit-multiflow/blob/21a0bee0f9565dafb9aa3b69f8206ee0c411a3f9/src/skmultiflow/trees/hoeffding_tree_regressor.py#L511-L543

I believe this last mentioned feature was inherited from `HoeffdingTreeClassifier`. We could possibly add a call to `remove_bad_splits` too. But in IMHO, that should not be done directly in `HoeffdingTreeRegressor`. Perhaps, in a common method to all the regression-based learning nodes. In this sense, the tree would not directly access the attribute observers, and each node would handle its memory management separately (when applicable).

What do you think?

jacobmontiel:
> I believe this last mentioned feature was inherited from HoeffdingTreeClassifier. We could possibly add a call to remove_bad_splits too. 

Pruning the tree is desirable given the potential efficiency boost. This has been explored both in classification and regression as you mention, so I believe we should have it.

> But in IMHO, that should not be done directly in `HoeffdingTreeRegressor`. Perhaps, in a common method to all the regression-based learning nodes. In this sense, the tree would not directly access the attribute observers, and each node would handle its memory management separately (when applicable).
> 
> What do you think?

I am having trouble understanding how is this going to be triggered if the functionality is moved to the nodes. Could you elaborate on this?

smastelini:
> I  am having trouble understanding how is this going to be triggered if the functionality is moved to the nodes. Could you elaborate on this?

I was thinking on calling a node's method, which, in turn, would access the attribute observers. The observers are private objects of the node, that's why I thought on avoiding accessing them directly from the tree.

smastelini:
> Pruning the tree is desirable given the potential efficiency boost. This has been explored both in classification and regression as you mention, so I believe we should have it.

I added the support for pruning the E-BST attribute observers. No changes were necessary on the trees' tests, but the ARFReg's ones had to be (slightly) modified.

Now, I have the sensation that the memory-management tests (related to #230) might have been impacted. Since the trees ought to use fewer memory resources now in the attribute observers, the memory-management tests might need to be updated to ensure the size control mechanisms are effectively triggered.

smastelini:
For completeness, I also ran some benchmarks comparing the original `HoeffdingTreeRegressor` (as in `master`) with the current version. This last one has a routine to remove potentially bad split candidates in the `NumericAttributeRegressionObserver` (as originally proposed in FIMT-DD's paper).

The following table summarizes the obtained results. No error measurements were included, since the resulting tree structures ought to be equivalent (in case of stationary streams):

image

PR #245, opened 2020-05-08
Fix bug in the LeveragingBaggingClassifier and NaiveBayes

### LINKED ISSUE #240: [BUG] Leveraging Bagging bug where input to drift detector is inverted.
**Describe the bug**
At line 287, the correct and incorrect classifications are reversed. At the moment, it is:
`add = 1 if (pred[0] == y_coded) else 0`
but should be:
`add = 0 if (pred[0] == y_coded) else 1`

In addition, while using Leverage Bagging with NaiveBayes base classifier, the reset of the classifier (line 305, `self.ensemble[i_max].reset()`) results in the deletion of class list and priors and the newly created (reset) classifier throws TypeError and returns 0 (line 375) while predicting. I have avoided this by calling partial_fit after reset, but not sure if this is the best practice.

jacobmontiel:

- Correct the value passed to ADWIN in `LeveragingBaggingClassifier` to monitor performance, previously this value was inverted.
- Add protection against division by zero.
- Update test cases and improve coverage.
- Remove dead code in `LeveragingBaggingClassifier`.
- Replace `adwin_ensemble` creation with list comprehension.
- Remove obsolete `try` in `LeveragingBaggingClassifier ` from the time when not all models were defaulting to prediction=[0] when empty. This is no longer necessary as now all classifiers return a valid value.
- Make the output of `predict_proba` in `NaiveBayes` 2D when the model is empty.
- Fixes #240 

---

#### Implementation

#### Tests

### DIFF
--- a/src/skmultiflow/meta/leverage_bagging.py
+++ b/src/skmultiflow/meta/leverage_bagging.py
@@ -171,9 +171,7 @@ class LeveragingBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorM
             self.base_estimator.reset()
         self.actual_n_estimators = self.n_estimators
         self.ensemble = [cp.deepcopy(self.base_estimator) for _ in range(self.actual_n_estimators)]
-        self.adwin_ensemble = []
-        for i in range(self.actual_n_estimators):
-            self.adwin_ensemble.append(ADWIN(self.delta))
+        self.adwin_ensemble = [ADWIN(self.delta) for _ in range(self.actual_n_estimators)]
         self._random_state = check_random_state(self.random_state)
         self.n_detected_changes = 0
         self.classes = None
@@ -325,14 +323,6 @@ class LeveragingBaggingClassifier(BaseSKMObject, ClassifierMixin, MetaEstimatorM
                 condition = ((n_ones - n_zeros) * (n_ones - n_zeros) >
                              (self.actual_n_estimators % 2))
         self.init_matrix_codes = False
-
-    def __adjust_ensemble_size(self):
-        if len(self.classes) != len(self.ensemble):
-            if len(self.classes) > len(self.ensemble):
-                for i in range(len(self.ensemble), len(self.classes)):
-                    self.ensemble.append(cp.deepcopy(self.base_estimator))
-                    self.adwin_ensemble.append(ADWIN(self.delta))
-                    self.actual_n_estimators += 1
 
     def predict(self, X):
         """ Predict classes for the passed data.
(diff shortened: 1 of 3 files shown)


src/skmultiflow/meta/leverage_bagging.py
```diff
@@ -244,7 +244,8 @@ def __partial_fit(self, X, y):
                     k = 1.0
                 elif pred[0] != y:
                     k = 1.0
-                elif self._random_state.rand() < (error / (1.0 - error)):
+                elif (error != 1.0 and
+                      self._random_state.rand() < (error / (1.0 - error))):
```
smastelini, line 248: Could you elaborate on the reason for the additional check (`error != 1.0`)?

src/skmultiflow/meta/leverage_bagging.py
```diff
@@ -281,17 +282,14 @@ def __partial_fit(self, X, y):
...
-                pred = self.ensemble[i].predict(np.asarray([X]))
-                if pred is not None:
-                    add = 1 if (pred[0] == y_coded) else 0
-                    error = self.adwin_ensemble[i].estimation
-                    self.adwin_ensemble[i].add_element(add)
-                    if self.adwin_ensemble[i].detected_change():
-                        if self.adwin_ensemble[i].estimation > error:
-                            change_detected = True
-            except ValueError:
-                change_detected = False
```
smastelini, line 294: Do not we expect this portion to throw a `ValueError` exception anymore? Could you elaborate on this change too?

src/skmultiflow/meta/leverage_bagging.py
jacobmontiel: This part of the code was unprotected and resulted in a `DivisionByZeroError` when `error==1.`

src/skmultiflow/meta/leverage_bagging.py
jacobmontiel: Sorry about that. This is an obsolete check from the time when not all models were defaulting to `prediction=[0]` when empty. This is no longer the case as now all `scikit-multiflow` classifiers return a valid value.

src/skmultiflow/meta/leverage_bagging.py
smastelini: Got it! Thank you for the answer!

src/skmultiflow/meta/leverage_bagging.py
smastelini: Thank you for the clarification!

PR #247, opened 2020-05-12
Fix PEP8 problems

smastelini:
Addresses two minor problems that are making fail the recently added linting check:

* Remove inadvisable usage of 'l' as variable name (E741) in `core.base`
* Fix missing string.format parameter (F523) in `data.DataStream`

---

#### Implementation

#### Tests

### DIFF
--- a/src/skmultiflow/data/data_stream.py
+++ b/src/skmultiflow/data/data_stream.py
@@ -403,7 +403,7 @@ class DataStream(Stream):
         if self.task_type == self._CLASSIFICATION:
             return "{}{} target(s), {} classes".format(name, self.n_targets, self.n_classes)
         elif self.task_type == self._REGRESSION:
-            return "{} target(s)".format(name, self.n_targets)
+            return "{}{} target(s)".format(name, self.n_targets)
 
     def _get_target_values(self):
         if self.task_type == 'classification':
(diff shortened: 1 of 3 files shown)


src/skmultiflow/data/data_stream.py
```diff
@@ -403,7 +403,7 @@ def get_data_info(self):
         if self.task_type == self._CLASSIFICATION:
             return "{}{} target(s), {} classes".format(name, self.n_targets, self.n_classes)
         elif self.task_type == self._REGRESSION:
-            return "{} target(s)".format(name, self.n_targets)
+            return "{}{} target(s)".format(name, self.n_targets)
```
garawalid, line 406: Don't we need some space here between the two **{}**?

smastelini: Hi @garawalid, thanks for pointing it out! I just checked the `name` variable definition:

https://github.com/scikit-multiflow/scikit-multiflow/blob/9c4168e37d0f638a698d0af2916aba866e83dbe8/src/skmultiflow/data/data_stream.py#L402

It already adds a space in case `self.name` is not `None` (`None` is the default value during the object creation).

garawalid: Great!

PR #248, opened 2020-05-15
Fix CI/CD pipeline and streamline sklearn-dependent tests

smastelini:
With the recent release of `sklearn==0.23` some of the tests in our CI/CD pipeline have broken. This problem happened because of the following change in sklearn, which is stated in the new version's changelog:

image

In simple terms, the string representation of a sklearn model only exhibits the parameters that were changed by the user, i.e., the ones that did not keep the suggested default values. 

Some of our automatic tests depend on sklearn and compare the skmultiflow models' descriptions with template strings. These strings previously depicted all the existing parameters of the sklearn models (which were usually used as base models for skmultiflow's meta models). This is no longer the case, as previously described.

We had two options:

1. Force sklearn to keep showing all the parameters and do not change the existing tests
2. Accept the new sklearn's behaviour and adapt our tests (ensuring backward compatibility with older versions of sklearn)

I opted to follow option 2 since it actually simplifies our tests and is future-proof (in case some parameters are renamed, for instance).

* Fix broken CI/CD pipeline
* Enables support to `sklearn==0.23`
* Enforce PEP8 style formatting in the changed files

---

#### Implementation
- [ ] Implementation is correct (it performs its intended function).
- [ ] Code is consistent with the framework.

#### Tests

### DIFF
--- a/tests/meta/test_batch_incremental.py
+++ b/tests/meta/test_batch_incremental.py
@@ -2,7 +2,11 @@ from skmultiflow.data import RandomTreeGenerator
 from skmultiflow.meta.batch_incremental import BatchIncrementalClassifier
 
 from sklearn.tree import DecisionTreeClassifier
+from sklearn import set_config
 import numpy as np
+# Force sklearn to show only the parameters whose default value have been changed when
+# printing an estimator (backwards compatibility with versions prior to sklearn==0.23)
+set_config(print_changed_only=True)
 
 
 def test_batch_incremental():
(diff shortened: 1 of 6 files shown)


smastelini:
I also suggest updating the `requirements.txt` file. I ran the automatic tests in my local machine using multiple versions of sklearn and I observed the following:

* CI/CD fails with `sklearn==0.20.4` (the current version indicated in the requirements file)
* Two tests fail with `sklearn==0.21.3`:
    - `tests/lazy/test_sam_knn.py::test_sam_knn_coverage`
    - `tests/neural_networks/test_perceptron.py::test_perceptron`
    -> in both cases the obtained results do not match with the expected ones.

If I recall correctly, some linear models and neighbour-based algorithms were impacted between versions `0.21` and `0.22` of sklearn.

jacobmontiel:
Changes are correct. Thanks for addressing this @smastelini

PR #256, opened 2020-06-14
Ifn dev

moravitan:

* Adding IFN base algorithm (single-target).
* Adding 4 IFN based algorithms supporting data-stream.

---

#### Implementation
- [X] Implementation is correct (it performs its intended function).
- [X] Code is consistent with the framework.
- [X] PR description covers ALL the changes performed.

#### Tests
- [X] New functionality is tested.
- [X] Tests are created for the new functionality or existing tests are updated accordingly.
- [X] ALL tests pass with no errors.

#### Verification Check
input: Iris dataset (from pandas), alpha = 0.95
output: 
network structure: 
TARGET ATTRIBUTE NO. 0 (Class)
0	1	2	3	4	
2	5	6		
3	7	8	9	
9	10	11	

*0 means root node.*
*each number represent a node.* 
*The first number in each line represent the node being splitted and the following numbers indicate its successors.*
*first layer split - PetLength (splited nodes in this layer - 0)*
*second layer split - PetWidth (splited nodes in this layer - 2,3)*
*third layer split - SepWidth (splited nodes in this layer - 9)*


src/skmultiflow/trees/__init__.py
```diff
@@ -20,12 +21,18 @@
...
            "HoeffdingTreeClassifier", "HoeffdingAdaptiveTreeClassifier", "ExtremelyFastDecisionTreeClassifier",
            "LabelCombinationHoeffdingTreeClassifier", "HoeffdingTreeRegressor", "HoeffdingAdaptiveTreeRegressor",
            "iSOUPTreeRegressor",
            "HoeffdingTree", "HAT", "HATT", "LCHT", "RegressionHoeffdingTree",
            "RegressionHAT", "MultiTargetRegressionHoeffdingTree",
-           "StackedSingleTargetHoeffdingTreeRegressor"]
-
+           "StackedSingleTargetHoeffdingTreeRegressor", "IfnClassifier",
+           "BasicIncremental", "MultipleModel", "PureMultiple",
+           "OnlineNetworkRegenerative"]
```
smastelini, line 38:
```suggestion
           "OnlineNetworkRegenerative"]

```

smastelini:
Hi everyone. Firstly, I would like to thank you for the vast amount of work you did so far! This new whole family of learning algorithms is going to be a nice addition to skmultiflow! There's really a lot of code to review, so as a first step, I did a scan of the changes looking for high-level/generic aspects that might not conform with the framework. As we progress with the changes, I will perform more "waves" of review and delve into implementation details. Please, do not hesitate to contact me if you have any questions.

I assembled some general points that deserve attention. For the most part, they can be solved with some refactoring and renaming some files/classes.

* Naming pattern:
    - I would change the prefix `Ifn` to `IFN`
    - Some of the proposed algorithms miss the 'Classifier' suffix, as suggested in the guidelines for contributors
    - In some cases, the name choice is too generic (considering that the classes are exported to the "outside world" in the `__init__.py` files)
        -> For example: `HiddenLayer`, `MetaLearning`. These names might have different interpretations depending on the context. My suggestion, again, is to use the IFN prefix as much as possible. Other suggestions are welcome too.
* At some point, a binary search is implemented. I suggest using the package `bisect` instead, which comes from the standard Python distribution. We might get a boost in the search performance.
* The `generate` interface used for most of the IOLIN-based IFN variants does not follow the `ClassifierMixin` standards
    - This part is one of the most problematic since all other algorithms available in the `trees` submodule follow the same pattern
    - Ideally, the IFN methods should not the IFNs handle their buffers internally, very much like the `BatchIncrementalClassifier`
        -> They would trigger model retraining and manage the sliding windows of examples in the `partial_fit` implementation
        -> In this case, they would use the currently active network to provide predictions when `predict/predict_proba` was called
        -> The structure depicted in Figure 2 of the original article would still be active, but it would act in the background, rather than acting as the primary access interface
    - Therefore, an instance of `Stream` would not be required anymore for the creation of an IFN

Some minor observations:

* Predictions should be integers (please, refer to the test files)
* Some tests point to the now gone datasets (in `src/skmultiflow/data/datasets`)
* In the `test_iolin` the function `_setup_eliminate_nodes_test_env` returns the assembled network but its value is never assigned to a variable. The network is accessed as a global variable
* There is no information about the `window_size` in `IfnClassifier`'s documentation

moravitan:
Following the given code review we made the following changes for now:
- adding IFN prefix to our files and classes
- adding use in the bisec package
- changing prediction to integers
- fixing the access to the global variable 'network' in `test_iolin`
- adding documentation to the 'window_size' field in `ifnClassifier`

Issue #219, opened 2020-02-26
[FEATURE] Add linter to CI/CD pipeline 


garawalid:
**Is your feature request related to a problem? Please describe.**
Actually, for each PR, the reviewer checks manually if the style PEP8 is correct. 

**Describe the solution you'd like**
It's possible to use `flake8` to check the PEP8 style. We integrate it into the CI/CD pipeline.

We still need to agree on some PEP8 rules:
- Do we keep max-line-length 120 (the same we do now) or we use 79 (as suggested by [PEP8](https://www.python.org/dev/peps/pep-0008/#id19))? 
- We need to ignore [E741](https://www.flake8rules.com/rules/E741.html) since we use a lot `x` and `y`.

garawalid:
Related to #213

garawalid:
I made a quick test and I found only these two style exceptions. Any thoughts?
@jacobmontiel @smastelini

jacobmontiel:
Thanks for creating the issue @garawalid 

My two cents:

* Max line length
   This is a hot topic, and most of the time goes down to team preferences [(E501)](https://www.flake8rules.com/rules/E501.html). Personally, I find 79 characters rather restrictive. We could consider a middle ground at 99. Additionally, we need to consider the number of lines to change as we consider smaller line lengths.

* CapitalizedWords or single upper case letter
   I understand what you mean. I am not sure if it fits into E741 though. In this case, we are following a scikit-learn convention that is familiar to a large user base so I vote to ignore this rule.

garawalid:
@jacobmontiel
I find 99 characters good as well, especially when I use a wide screen. 

I'll create an initial PR ASAP.

smastelini:
Hi @garawalid and @jacobmontiel. I use a flake8-based plugin in my text editor, and I always received many warnings regarding the line lengths.  I also agree that 99 characters are a good compromise for our case.

@garawalid, this upcoming PR is going to be an excellent addition for the CI/CD pipeline! Thanks for that :)

garawalid:
Using a 99-character rule will lead to many changes. I made a PR (#221) with an initial linting.  
I've been thinking of using 120 characters. What do you think?

jacobmontiel:
> Using a 99-character rule will lead to many changes. I made a PR (#221) with an initial linting.
> I've been thinking of using 120 characters. What do you think?

[...]

PR #262, opened 2020-06-30
Lint drift detection part

garawalid:
Part of #219


src/skmultiflow/drift_detection/kswin.py
```diff
@@ -89,7 +91,7 @@ def __init__(self, alpha=0.005, window_size=100, stat_size=30, data=None):
         if self.window_size < self.stat_size:
             raise ValueError("stat_size must be smaller than window_size")
 
-        if type(data) != np.ndarray or type(data) is None:
+        if not isinstance(data, np.ndarray) or isinstance(data, None):
```
smastelini, line 94:
```suggestion
        if not isinstance(data, np.ndarray) or data is None:
```

This could be the culprit making the tests fail.

smastelini:
Hi @garawalid. Thank you for keeping the linting efforts! I left a small comment that might address the cause why the tests are failing.

src/skmultiflow/drift_detection/kswin.py
garawalid: You are right!

garawalid:
@smastelini thanks for the review! I need also to fix some regression in the lint in `src/skmultiflow/data`

garawalid:
Something wrong with flake8 and `src/skmultiflow/data`. In fact, the pipeline passes in the last commit in the master (Also on my local machine). And I didn't change it here.

garawalid:
In fact, there is a bug in flake8 (pycodestyle) with W605 and W503. I reported it here PyCQA/pycodestyle#949. I'll disable both rules for instance!

garawalid:
It was a mistake in the `tox.ini` file and I fixed it! This PR is ready for review.

smastelini:
Thank you for the hard work @garawalid. I'll proceed to review the PR.

src/skmultiflow/drift_detection/ddm.py
```diff
@@ -61,17 +61,19 @@ class DDM(BaseDriftDetector):
...
     >>> # increase in error rate
     >>> for i in range(999, 1500):
     ...     data_stream[i] = 0
     >>> # Adding stream elements to DDM and verifying if drift occurred
     >>> for i in range(2000):
     ...     ddm.add_element(data_stream[i])
     ...     if ddm.detected_warning_zone():
-    ...         print('Warning zone has been detected in data: ' + str(data_stream[i]) + ' - of index: ' + str(i))
+    ...         print("Warning zone has been detected in data: {} - of index: {}"
+    ...                 .format(data_stream[i],i))
```
smastelini, line 73:
```suggestion
    ...         print("Warning zone has been detected in data: {} - of index: {}"
    ...                 .format(data_stream[i],i))
```

smastelini, line 73: The number of spaces before the the `.format` statement seems to be incorrect. Nonetheless, when I try to create a suggestion, the indentation looks correct in the code editing tool. Visually, the format command should be aligned with the start of the string. Perhaps the rendering process is making it look unaligned. Any thoughts @garawalid?vvvvvvvvvvv

src/skmultiflow/drift_detection/ddm.py
```diff
@@ -139,10 +141,12 @@ def add_element(self, prediction):
             self.miss_sd_min = self.miss_std
             self.miss_prob_sd_min = self.miss_prob + self.miss_std
 
-        if self.miss_prob + self.miss_std > self.miss_prob_min + self.out_control_level * self.miss_sd_min:
+        if self.miss_prob + self.miss_std > self.miss_prob_min + self.out_control_level \
+                * self.miss_sd_min:
```
smastelini, line 145: This line break violates W503.

src/skmultiflow/drift_detection/ddm.py
```diff
@@ -139,10 +141,12 @@ def add_element(self, prediction):
...
             self.in_concept_change = True
 
-        elif self.miss_prob + self.miss_std > self.miss_prob_min + self.warning_level * self.miss_sd_min:
+        elif self.miss_prob + self.miss_std > self.miss_prob_min + self.warning_level \
+                * self.miss_sd_min:
```
smastelini, line 149: This line break violates W503.

src/skmultiflow/drift_detection/eddm.py
```diff
@@ -47,18 +47,19 @@ class EDDM(BaseDriftDetector):
...
     >>> # increase in error rate
     >>> for i in range(999, 1500):
     ...     data_stream[i] = 0
     >>> # Adding stream elements to EDDM and verifying if drift occurred
     >>> for i in range(2000):
     ...     eddm.add_element(data_stream[i])
     ...     if eddm.detected_warning_zone():
-    ...         print('Warning zone has been detected in data: ' + str(data_stream[i]) + ' - of index: ' + str(i))
+    ...         print("Warning zone has been detected in data: {} - of index: {}"
+    ...                 .format(data_stream[i], i))
```
smastelini, line 59: The same indentation problem happens here.

src/skmultiflow/drift_detection/hddm_a.py
```diff
@@ -1,4 +1,4 @@
-from math import *
+from math import log, sqrt
```
smastelini, line 1: Nice catch! :D

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -131,16 +139,21 @@ def add_element(self, prediction):
     def _detect_mean_increment(self, sample1, sample2, confidence):
         if sample1.EWMA_estimator < 0 or sample2.EWMA_estimator < 0:
             return False
-
-        bound = sqrt((sample1.independent_bounded_condition_sum
-                      + sample2.independent_bounded_condition_sum) * log(1 / confidence) / 2)
+        sum = sample1.independent_bounded_condition_sum + sample2.independent_bounded_condition_sum
```
smastelini, line 142: `sum` is a reserved keyword in Python. I suggest changing the variable name. Appending a simple underscore after the word is enough: `sum_` (though I am not sure it is the best practice/workaround).

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -171,18 +187,21 @@ def _update_decr_statistics(self, value, confidence):
...
             else:
                 self.sample2_decr_monitor.EWMA_estimator = \
-                    self.lambda_option * value + aux_decay * self.sample2_decr_monitor.EWMA_estimator
+                    self.lambda_option * value + aux_decay * \
+                    self.sample2_decr_monitor.EWMA_estimator
                 self.sample2_decr_monitor.independent_bounded_condition_sum = \
                     self.lambda_option * self.lambda_option \
-                    + aux_decay * aux_decay * self.sample2_decr_monitor.independent_bounded_condition_sum
+                    + aux_decay * aux_decay \
+                    * self.sample2_decr_monitor.independent_bounded_condition_sum
```
smastelini, line 204: This line break violates W503.

smastelini, line 204: These line breaks violate W503.

src/skmultiflow/drift_detection/adwin.py
```diff
@@ -328,10 +335,13 @@ def detected_change(self):
...
                             break
 
-                        abs_value = 1. * ((u0/n0) - (u1/n1))
-                        if (n1 >= self.mint_min_window_length) and (n0 >= self.mint_min_window_length)\
-                                and (self.__bln_cut_expression(n0, n1, u0, u1, v0, v1, abs_value, self.delta)):
-                            bln_bucket_deleted = True
+                        abs_value = 1. * ((u0 / n0) - (u1 / n1))
+                        if (n1 >= self.mint_min_window_length) \
+                                and (n0 >= self.mint_min_window_length) \
+                                and (
```
smastelini, line 341: These lines violate W503.

src/skmultiflow/data/time_manager.py
```diff
@@ -1,156 +1,156 @@
...
-        if self._USE_SAMPLE_WEIGHT:
-            sample_weight = np.array(samples["weight"].values.tolist())
-        else:
-            sample_weight = None
-        # remove samples from queue
-        self._cleanup_samples(batch_size)
-        # return X, y_true, y_pred, and weight for the dequeued samples
-        return X, y_true, y_pred, sample_weight
+import numpy as np
+import pandas as pd
```
smastelini, line 2: Does this entire file was added without changes?

garawalid:
Thanks @smastelini for the review.  I think your `tox.ini` uses `ignore` instead of `extend-ignore`.  
The **W503** is disabled by default. See the full list [here](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore) and [here](https://gitlab.com/pycqa/flake8/-/blob/a7be77f761a4c29121d6bb6f61c11902281f9105/src/flake8/defaults.py#L15).

src/skmultiflow/data/time_manager.py
garawalid: I accidentally changed the file encoding from `CRLF` to `LF`. I deleted only one white space. Let me revert the changes.

garawalid: It's done! Now you can see the difference correctly.

smastelini: Perfect!

smastelini:
> Thanks @smastelini for the review. I think your tox.ini uses ignore instead of extend-ignore.
The W503 is disabled by default. See the full list here and here.

Thanks for the explanation and for pointing me out to these guidelines, @garawalid. In fact, I found these W503 violations visually :-)

Although W503 is ignored by default, (and only for readability and aesthetic purposes) I personally would prefer to enforce this rule. But it is worth discussing further. Any thoughts on that matter, @jacobmontiel?

garawalid:
@smastelini 
The W503 will be an anti-pattern, that's why they introduced the W504 (the opposite).
> Note: Despite being in the anti-pattern section, this will soon be considered the best practice.

I ve been reading some comments about W503 & W504. And it sounds that the community didn't reach a final agreement yet.
I'll vote to disable them both (the default configuration of flake8).
[Correct style for line breaks when chaining methods in Python](https://stackoverflow.com/questions/7942586/correct-style-for-line-breaks-when-chaining-methods-in-python)

smastelini:
Hi @garawalid, you convinced  me hahaha.

I would say, let's go for W504. What do you think @jacobmontiel?

src/skmultiflow/drift_detection/ddm.py
garawalid: In fact, it's me who forced the format to be in the second line, so I can respect the max length rule. In the IDE, it's working like a charm. I'll change it as you suggested.

src/skmultiflow/drift_detection/hddm_w.py
garawalid: Nice catch!

garawalid:
@smastelini  sorry for the late response, I missed some comments.
Could you please take another look whenever you have some free time?

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -113,7 +113,7 @@ def add_element(self, prediction):
             self.total.independent_bounded_condition_sum = 1
         else:
             self.total.EWMA_estimator = self.lambda_option * \
-                prediction + aux_decay_rate * self.total.EWMA_estimator
+                                        prediction + aux_decay_rate * self.total.EWMA_estimator
```
smastelini, line 116: Is there a reason to adopt this indentation style? Perhaps some "PEP*"? The previous version looked good to me, but I am not sure whether or not this new style is the preferred one according to community guidelines.

src/skmultiflow/drift_detection/hddm_w.py
```diff
@@ -173,8 +174,8 @@ def _update_incr_statistics(self, value, confidence):
                 self.sample2_incr_monitor.independent_bounded_condition_sum = 1
             else:
                 self.sample2_incr_monitor.EWMA_estimator = self.lambda_option * \
-                    value + aux_decay * \
-                    self.sample2_incr_monitor.EWMA_estimator
+                                                           value + aux_decay * \
+                                                           self.sample2_incr_monitor.EWMA_estimator
```
smastelini, line 178: Same here

smastelini:
Hi @garawalid, thanks for the changes. I left some additional comments and resolved some of the pending items that were already solved.

src/skmultiflow/drift_detection/hddm_w.py
smastelini: :D

src/skmultiflow/drift_detection/ddm.py
smastelini: Thanks for addressing that!

src/skmultiflow/drift_detection/hddm_w.py
garawalid: I used [autopep8](https://github.com/hhatto/autopep8) to format the file. Do you think it's an issue?

smastelini: I don't think so. As long as the linting tests pass, I'm cool with that! :D

smastelini:
Thanks Walid for your linting efforts! This PR is now ready to be merged.

garawalid:
Thanks @smastelini for your time and the review!

PR #264, opened 2020-07-06
Add missing 'has_more_samples' method to MultiLabelGenerator

### LINKED ISSUE #261: [BUG] Method "has_more_samples" from data.MultilabelGenerator always returns True
**Describe the bug**
The Multi-label stream generator class does not implement the method "has_more_samples". The parent class does, but it always returns True. This is particularly problematic when this stream generator is used within the ConceptDriftStream class, because the method is called internally. 

**To Reproduce**
```
multilabel_stream = MultilabelGenerator(n_samples=10)
multilabel_stream.next_sample(10)
multilabel_stream.has_more_samples()
```
Returns "True".

**Expected behavior**
It should return False whenever the stream is exhausted.

smastelini:
Add missing `has_more_samples` method to `MultiLabelGenerator`.

* Fixes #261

---

#### Implementation

#### Tests


garawalid (approved):
LGTM!

Issue #170, opened 2019-12-02
[FEATURE] Bootstrap sampling option for the Hoeffding Adaptive Tree Regressor


jacobmontiel:
By using bootstrap sampling at the leaf nodes we can improve the performance of the Hoeffding Adaptive Tree classifier. This is not currently available in the Hoeffding Adaptive Tree Regressor.

This issue is intended to open the discussion on the applicability and implementation (if required) of this feature. A similar solution to the one in #169 is expected.

Issue #157, opened 2019-11-07
[ENHANCEMENT] Check whether there are redundancies among the tree learning nodes


smastelini:
PR #148 separated the learning nodes implementations from the tree algorithms.

As leaf predictors, these nodes have a similar function and follow similar strategies to provide responses. Therefore, there might be redundancies among their code, which could be avoided with proper class inheritance usage.

This is something worth investigating. Besides removing unnecessary portions of the code, these changes would also make easier the framework maintenance and update.

PR #271, opened 2020-07-21
Tree module refactor

### LINKED ISSUE #156: [ENHANCEMENT] Enforce naming convention for the trees' learning nodes and add proper documentation
This Issue in some sense relates to #138.

PR #148 separated the learning node implementations from the Hoeffding Trees source codes.
Nonetheless, this exposed the fact that these predictors do not follow a pattern in their naming, nor they have proper documentation for all their methods.

A useful enhancement would be adopting a clear, concise, and simple naming convention. This convention should indicate the type of learning node (Active or Inactive), the target learning task (Classification or Regression), and the type of the predictor (Naive Bayes, Perceptron, and so forth). Other meaningful information could also be added to their names.

Those changes ought to impact most of the learning node implementations. So they may result in the necessity of triggering warnings for the users and maintaining the current naming during some upcoming versions of skmultiflow.

Suggestions are very much welcomed!

smastelini:
This PR addresses a few open issues. Its main goal is to simplify the underlying structure of the tree nodes and remove duplicated code.

## Context

The tree algorithms in skmultiflow operate by manipulating basically two kinds of nodes:
* Split Nodes (a.k.a. decision nodes): The ones used to sort incoming instances down the tree
* Learning Nodes (typically leaves): The nodes that provide responses and learn from examples

As the tree monitor more examples, leaves might become splits, which in turn will have additional leaves and so on. 

Most of the Split nodes do not perform learning operations, however, there are some exceptions such as the `HoeffdingAdaptiveTree*` variants.

Moreover, the typical learning nodes, i.e., leaves, can be of two types:

* Active: monitor the relation between incoming features and the target variable in order to perform splits and enable tree growth
* Inactive: only update basic statics (to provide responses), but do not support tree growth

Active leaves are the default in the trees, but they might be turned into Inactive ones to save memory (the `max_byte_size` and `mem_check_period` parameters control the memory management). In summary, Active and Inactive Leaves are practically the same, except that the latter type does not carry attribute observers (please check `trees/attribute_observers` for more information).

## Problem

In the current implementation the learning operations of Active and Inactive nodes are repeated multiple times. These operations 
could be implemented in a single place and be inherited to multiple nodes

## Objective

This PR addresses the mentioned problem by simplifying the way in which the tree nodes are structured. We introduce the notion of base `ActiveLeaf` and `InactiveLeaf` classes, and also improve the current implementation of the base `LearningNode`. By doing so, we simplify the implementation of the actual nodes used in the tree algorithms.

The core operations are not repeated anymore but inherited from a common implementation across multiple nodes. These changes make the process of creating or extending existing tree nodes much easier! They also facilitate the maintenance of the code.

---

* Refactor the tree nodes 
* Place all nodes belonging to the same tree algorithm in the same file and standardize the file naming convention
* Enforce naming convention for the nodes
* Remove redundancies by applying a new tree node architecture
* Enforce PEP8 in the tree submodule (#219)
* Fixes #156, #157, and #170
* Fix bug in `HoeffdingAdaptiveTreeRegressor` and update strategy of this tree to monitor drifts (now the same approach applied in `AdaptiveRandomForestRegressor` is used)
* Fix bug in `ARFHoeffdingTreeRegressor` to ensure its consistency with `HoeffdingTreeRegressor`

---

#### Implementation

#### Tests


MaxHalford:
Saving 553 lines of code is nice :)

smastelini:
Some portions of the `HoeffdingAdaptiveTreeRegressor` are not reached because we temporarily disabled the usage of "option predictions". In `HoeffdingAdaptiveTreeClassifier` when predicting new data the tree sorts incoming instances to both the main structure and also possible existing alternate trees. The final responses are taken by summing all the reached leaves' statistics and normalizing them. That is what I previously referred as "option predictions" as this strategy bears similarities with Option Decision Trees.

In regression option predictions should be implemented differently since the leaves provide the the final responses directly instead of yielding classes probabilties. Hence, the previous implementation was not completely correct as it only encompassed the case `leaf_prediction='mean`.

We should create another PR to handle this specific situation.
The unreachable code can either be maintained for future reference or removed in this PR so that we increase coverage.

garawalid:
Nice work @smastelini !
I have some comments about how we structure the `node` module. I agree that there are many classes of nodes and each one is a separate Python file, But I still find the class name in file is more clear than grouping many classes in one file.

For example: 
`RandomActiveLeafClass`, `RandomActiveLearningNodeMC`, `RandomActiveLearningNodeNB`, `RandomActiveLearningNodeNBA` are in `_arf_htc_nodes.py`

I suggest that we keep one file for one class and we group them with packages following the same semantic. If I take the same example, it will be as follows:
/nodes/random/random_active_leaf.py -> RandomActiveLeaf

src/skmultiflow/trees/extremely_fast_decision_tree.py
```diff
@@ -340,21 +340,20 @@ def _sort_instance_into_leaf(self, X, y, weight):
             found_node.parent.set_child(found_node.parent_branch, leaf_node)
             self._active_leaf_node_cnt += 1
 
-        if isinstance(leaf_node, LearningNode):
+        if not isinstance(leaf_node, SplitNode):
```
garawalid, line 343: If we add a new type of Node, then it will pass the condition and we won't find necessary the learn() method. 
I'd keep it like before.

jacobmontiel: The new `EFDTSplitNode` extends the `LearningNodeMC` which in place implements the `learn_one` method.

In this case, we need to change `EFDTSplitNode` to explicitly implement `learn_one` and not inherit it.

Does that sound good @garawalid @smastelini ?

smastelini: I agree with you both. I changed the test because it would trigger a training in the split node in this new organization. But if we remove the inheritance from `LearningNodeMC` we can keep the previous test, as better depicts what is really happening.

smastelini: In commit [ff02f6](https://github.com/scikit-multiflow/scikit-multiflow/pull/271/commits/ff02f65e5ec5e80bcfe17dfbe19b03b2ea13813e) I made the `EFDTSplitNode` implement the needed methods without inheriting from `LearningNodeMC`. Then I reverted the check as it was previously, as suggested by @garawalid.

garawalid: @jacobmontiel I agree with you!
@smastelini that's great!

smastelini:
> I have some comments about how we structure the `node` module. I agree that there are many classes of nodes and each one is a separate Python file, But I still find the class name in file is more clear than grouping many classes in one file. 
> For example:
> `RandomActiveLeafClass`, `RandomActiveLearningNodeMC`, `RandomActiveLearningNodeNB`, `RandomActiveLearningNodeNBA` are in `_arf_htc_nodes.py`

Hi @garawalid, sorry for my late response. That's indeed a valid concern. Yesterday we have finally covered one last change in this PR: we made the sub-modules of `trees` private. All of these components (attribute observers, split tests, tree nodes, etc.) are neither intended to be used standalone nor by the end users. They only make sense as building blocks of the tree algorithms. So, by making them private we comply with [PEP8](https://www.python.org/dev/peps/pep-0008/?#public-and-internal-interfaces). 

The idea behind grouping the nodes belonging to the same tree in a file is to facilitate future maintenance and also to make it clear to which algorithm each node belongs to. Going even further, I also removed the export of standalone `ActiveLeaf*` and `LearningNode*` from `_nodes.__init__.py`, since they are not used directly in the trees, only internally.

There might be still some work to do regarding the new importing scheme. Any suggestions are welcomed!

smastelini:
> -1400 lines of code approx., this is a great and welcomed improvement.
> Thanks @smastelini !

I am quite happy with the results we got so far! Thanks @jacobmontiel and @garawalid for the great suggestions!

> The changes are correct. I just found some TODO tags in the code that can be removed as they are no longer valid. Once they are removed we can close the PR.

Nice catch! I removed the outdated TODO tags in commit [dbac08d](https://github.com/scikit-multiflow/scikit-multiflow/pull/271/commits/dbac08d46a89ff5f4086bf6a9847d1e03da09719).

garawalid:
@smastelini 
I'm convinced of the grouping semantic of nodes per module. 
I'd make the `_nodes` package without the underscore like this `nodes`. As you said it's a private package also it sounds PEP8 discouraging this kind of naming.
Package and Module Names](https://www.python.org/dev/peps/pep-0008/#package-and-module-names)
> Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.
> When an extension module written in C or C++ has an accompanying Python module that provides a higher level (e.g. more object oriented) interface, the C/C++ module has a leading underscore (e.g. _socket).

Also, I'd add a note to `CONTRIBUTING.md` explaining the structure of `trees` package, the semantic of `nodes` and any useful note. I will make it easy for new contributors.

tests/meta/test_adaptive_random_forest_regressor.py
```diff
@@ -116,9 +116,9 @@ def test_adaptive_random_forest_regressor_perceptron():
     error2 = mean_absolute_error(y_true, y_pred2)
     error3 = mean_absolute_error(y_true, y_pred3)
 
-    expected_error1 = 118.69
-    expected_error2 = 121.56
-    expected_error3 = 117.96
+    expected_error1 = 126.01
```
garawalid, line 119: I'm wondering why the expected error changed?

tests/trees/test_hoeffding_adaptive_tree_regressor.py
```diff
@@ -246,9 +238,9 @@ def test_hoeffding_adaptive_tree_regressor_alternate_tree():
         cnt += 1
 
     # Root node changed
-    expected_info = "if Attribute 1 <= -0.00015267114158334927:\n" \
-        "  Leaf = Statistics {0: 904.0000, 1: 1098.6423, 2: 332597.7050}\n" \
-        "if Attribute 1 > -0.00015267114158334927:\n" \
-        "  Leaf = Statistics {0: 905.0000, 1: 17227.8522, 2: 332000.7548}\n"
+    expected_info = "if Attribute 1 <= 0.02469103490619995:\n" \
```
garawalid, line 241: The same goes here!

smastelini:
> @smastelini
> I'm convinced of the grouping semantic of nodes per module.

:D

> I'd make the `_nodes` package without the underscore like this `nodes`. As you said it's a private package also it sounds PEP8 discouraging this kind of naming.
> Package and Module Names](https://www.python.org/dev/peps/pep-0008/#package-and-module-names)
> 
> > Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.
> > When an extension module written in C or C++ has an accompanying Python module that provides a higher level (e.g. more object oriented) interface, the C/C++ module has a leading underscore (e.g. _socket).

It could be just my interpretation of this, but I think that underscores are discouraged in the case of long modules names, e.g. `long_module_name_with_underscores`. In this case, we are using a single leading underscore to indicate the internal purpose of the modules.

> Also, I'd add a note to `CONTRIBUTING.md` explaining the structure of `trees` package, the semantic of `nodes` and any useful note. I will make it easy for new contributors.

You are totally right! This a *must* to facilitate future collaboration. I believe the tree module structure is somewhat future-proof as it is now. So we can come up with such documentation. What do you think @jacobmontiel?

tests/meta/test_adaptive_random_forest_regressor.py
smastelini: While simplifying the node structure we found a minor bug in the `ARFHoeffdingTreeRegressor` nodes. That's why the results slightly changed

jacobmontiel:
I think that `_nodes` achieves the objective of indicating that the content of this sub-package if for internal usage only. 

Regarding the notes in `CONTRIBUTING.md` this should be approach as a _generic_ best practice. This is something that we want to enforce on all modules and packages. We can use the `trees` subpackage as an example. It is very unlikely that somebody would follow a different pattern (not impossible though).

tests/meta/test_adaptive_random_forest_regressor.py
garawalid: I see, thanks @smastelini

garawalid:
Okay, let's keep the name of the nodes package this way!

jacobmontiel:
Awesome work @smastelini !
I am closing and merging this PR.

PR #276, opened 2020-09-25
Create developer tools, to ease setup and development

jmrozanec:

* We add a script (multiflow.sh) with developer utilities. 
* We provide a dockerized environment for scikit-multiflow development, mirroring local code into a Docker instance. 
* The developer no longer requires to install any dependencies and can execute tests from a Docker, bringing the same dev experience to all contributors.
* We provide documentation on how these tools can be used in DEVELOPMENT.md.
* Requirements on developers PC: Docker, Python, pip, npm

We think such tools are important, since lowering the barrier to start developing on scikit-multiflow, encourages more people to get involved in the project and fosters contribution and adoption.

---

#### Implementation

#### Tests


jmrozanec:
@jacobmontiel please let us know if anything else is required for this PR acceptance. Thanks!

jmrozanec:
@jacobmontiel is there anything we can do to help to accept this PR?

garawalid:
@jmrozanec  Thanks for the PR!
I'll review this PR tonight!

garawalid:
Thanks @jmrozanec for your work!

We already have a [scikit-multiflow docker image](https://github.com/scikit-multiflow/scikit-multiflow/tree/master/docker) with a repository. The one with `devel` tag tracks changes of each commit in the master branch.

It's good to use docker as a development environment. But I don't agree with putting it in the repository for many reasons:

- We cannot force users to use docker as a development environment. (I prefer using the OS directly because it's faster)
- The cost of maintenance is expensive.
- The development environment is unique to each person, for example, I don't see any reason to use `tmux`.

If you have some improvement in the docker file or in the DevOps pipelines feel free to open a PR!

jmrozanec:
@garawalid we view it in a different way: this is not forcing anyone to use the tools, it just provides means to do so quicker. If you check the current image Vs what we provided, there is a difference on why is provided. We would suggest to converge into a single version, but accept this approach for this, who want it. 
In this case, tmux is optional as well.

garawalid:
> It just provides means to do so quicker: the tools are an additional option for devs. 

I'm not sure it's quicker, getting started is made in two steps:
1. clone the Repository
2. Install requirements

And there are many possible ways to install requirements (on virtual env. or the host python)

> If you check the current image vs what we provided, there is a difference in why is provided. Our image does not freeze any code or dependencies: just mirrors the repo code in docker, so you avoid installing any dependencies and are ready to develop. Thus the cost of maintenance is zero.

- In theory, the cost of maintenance is zero but it's beyond that! In my company, we have a docker env. and we faced many cases where we need to update the docker env and build the docker from the scratch. A real use case: A contributor adds a new python dependency that needs some os packages (should be installed with apt for example) and for some reason, this contributor is not using the docker env. So, we need to manually update dockerfile with this package, otherwise, the env will be useless. This can lead to technical debt. I prefer that each developer handle his OS.

- Assume that I'm using the docker env. In order to use my IDE (Pycharm), I can either use mapping, remote interpreters or both.
  - Mapping a path: With Linux it's working like a charm, but I faced many issues with docker in Windows. More complexity to be handled. With Mapping alone, we cannot benefit from the full power of Pycharm (Debugger, run tests and generate coverage, profiling, etc)
  - Remote Python interpreter with ssh: This can be a good solution with the automatic upload. Developer can face some issues related to ssh configuration and network.
   
- The good point is the docker env. will isolate the dev. env. from the OS. But this can be done easily with `venv`. right now we don't have any big dependencies and extra services.

N.B: 
- We are assuming that anyone will use the docker env. has a good experience with Docker and Docker Compose.
- Some features of Pycharm are available only in the **PRO** version.

> We may converge into a single Docker version, but accept this approach for those, who may want it.
In this case, tmux is optional as well - does not force anyone to use it, even if using the multiflow.sh utilities.

As for velocity, the dockerized environment is fast and pretty standard nowadays as development env in many enterprise projects.

The [devel](https://github.com/scikit-multiflow/scikit-multiflow/blob/master/docker/dockerfiles/devel.Dockerfile) tag contains a clone of the repository and it's updated each time we commit the master. So technically, it's an image that contains a dev version of scikit-multiflow and we can use it in the development. Maybe we can add more dev. tools.

The whole point is to discuss the proposal and weigh the pros and cons. I think it's faster and easier to clone the project, install dependencies and start developing with less complexity especially when we face some bugs.

And I think the docker env. is a good solution for some project that has many dependencies (Hadoop ecosystem, Kafka or a Metastore). But for a python package that can be set up in two steps regardless of the OS is not worth it! 

I hope you are convinced! I vote -1 for this feature.

cc @jacobmontiel , @smastelini Any thoughts about this proposal?

jmrozanec:
@garawalid thanks - we appreciate the comments, and understand the -1. We agree for some may be easier to just perform two steps - it is ok, and will submit another PR with tools to do just that 😄  Just want to point out two things, that are not accurate:
  - in our case is zero maintenance, because packages are installed on the fly from the requirements and requirements-dev files. No dependencies are freezed in the Docker image, so will never be stale. Same with code: we mirror the local code, not commits in remote repo. This makes the difference.
- we may be soon integrating with Kafka and others, and will require this envs to test and show multiflow works ok with them. This provides an easy way to do so. We may choose to handle integrations in another repo, etc. as well - that may be a valid option (integrations may have different iteration velocities than main code). But will most likely require something similar, where integrations will be developed.

garawalid:
> we may be soon integrating with Kafka and others, and will require this envs to test and show multiflow works ok with them. This provides an easy way to do so. We may choose to handle integrations in another repo, etc. as well - that may be a valid option (integrations may have different iteration velocities than main code). But will most likely require something similar, where integrations will be developed.  

I agree, we need to figure out something ready to use for developers with zero maintenance cost. Most probably we will use docker env.

jmrozanec:
@garawalid ok, this tools provide it 😄 In the Kafka PR, we upgrade them, to provide Kafka environment as well.

Issue #106, opened 2019-04-13
[FEATURE] streamz as an input


martindurant:
I came across this work at anacondacon, and am impressed at the set of iterative learning methods implemented. However, I feel that the *input* data streams are not something that has so far been particularly important for you in development.

The pure-python real-time events processing library [streamz](https://streamz.readthedocs.io/en/latest/) may be just the link that can stand between the data sources (text files, TCP/HTTP server, kafka, times random processes ...), doing some data cleaning and processing, and then passing the events into the algorithms developed here. Indeed, the library comes with hooks for real-time automatic plotting, which might suit your project very nicely on the output side too.

This issue is here to investigate to what extent the two projects might collaborate for mutual benefit.

jacobmontiel:
Thanks for the information @martindurant

As you mention, data ingestion is something that we are looking forward to improving. I talked with some people at AnacondaCON regarding this topic and they also pointed to `streamz`. This is definitely worth investigating.

Could you elaborate on
> Indeed, the library comes with hooks for real-time automatic plotting

as we are also interested in improving in terms of visualization. I also talked with the guys involved in `pyviz` and I believe we are into something good here.

@jmread @abifet

martindurant:
As far as visualisation goes, I would firstly point you to the following page of examples
https://hvplot.pyviz.org/user_guide/Streaming.html

With the advent of ["Panel"](https://panel.pyviz.org/), a widget-interaction and feedback library, you can now embed such output in complete dashboard layouts, and serve these as stand-alone apps or in-notebook.

martindurant:
Some more examples using holoviews directly: http://holoviews.org/user_guide/Streaming_Data.html (the viz stack goes, low-level to high, bokeh-holoviews-hvplot-panel)

jacobmontiel:
Those are really interesting examples and thanks for the viz stack explanation, it can be confusing at first. I think `Panel` is worth considering if it allows to easily display related information and (optionally) configure what to display.

A minor "concern" is that holoviews uses web pages to display information and some users might find this as an overhead. Although this might be just my own perception.

Nonetheless, I think we can focus first on `streamz` and later on visualization, but it is good to know the potential.

martindurant:
[...]

PR #278, opened 2020-10-02
Support Kafka integration. Decouple evaluation and reporting through observers. 

jmrozanec:

* We support Kafka as events source
* We propose using data-sources to listen to events from any source. Data sources provide a stream of events, which we observe to train models and evaluate them based on different criteria (prequential, holdout (quantity and time based))
* We decouple train metrics computation from how are reported: this way, we can integrate with 3rd party tools to visualize metrics, persist/broadcast them where and when required, etc. We are no longer limited to the dashboard, file, and summarized info provided. 
* We introduce the mockito mocking library for tests to reduce the coupling of unnecessary elements for tests.
* We provide Dockerized environment for development, installing all required dependencies. This reduces setup requirements for developers, who can start working right away. The setup includes a docker instance with Kafka so that the Kafka integration can be verified/used.

---

#### Implementation
- [ ] PR description covers ALL the changes performed.

#### Tests


garawalid:
@jmrozanec thanks for this PR and the effort! 
This PR is big and difficult to review. Could you please divide it into small ones? You can keep the Docker environment the last one!  

Also, I think we need to discuss the design of the new API. What do you think?

jmrozanec:
@garawalid it is difficult to split the change into smaller ones, at least for the Kafka support, since changes are all related. We already split the dev tools into #276, which is awaiting acceptance.

garawalid:
@jmrozanec 
I'll review this PR ASAP. 

> Seems that in this PR, gh-actions do not detect the package structure changed, and fail due to that. Locally all tests pass. How can this be adjusted on PR checks?  

Do you run tests with an IDE (Pycharm for example) or directly with the command line?

jmrozanec:
@garawalid we run them directly from the command line

ahmed-masud:
Why not pull kafka through streamz? as suggested in https://github.com/scikit-multiflow/scikit-multiflow/issues/106

jmrozanec:
> Why not pull kafka through streamz? as suggested in #106

@ahmed-masud because we already developed the feature before getting to know the #106 issue, and #106 was left half done - seems none is working on it or having the time to push it forward. We would be glad to work towards an implementation that can support integrations through and outside streamz as well.

Issue #272, opened 2020-08-05
[BUG] SAMKNNClassifier cannot be reset


ginop:
**Describe the bug**
The `skmultiflow.lazy` class `SAMKNNClassifier` throws a non-obvious error when using the reset method.

> TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'

This appears to stem from the `__init__()` command generated via a string in which the keyword argument `max_window_size` is set to `None` rather than to the currently used value.

**To Reproduce**
```
from skmultiflow.lazy import SAMKNNClassifier
SAMKNNClassifier().reset()
```

**Expected behavior**
The SAMKNNClassifier should reset to untrained without changing its configuration.

jacobmontiel:
Thanks for reporting this @ginop 

The bug comes from a typo in the internal variable name:

```python
self.max_wind_size = max_window_size
```

The `reset` method fails to set the `max_window_size` resulting in the mentioned error.

PR #281, opened 2020-10-28
Fix sam_knn.py variable name for reset

ginop:
As presented here, https://github.com/scikit-multiflow/scikit-multiflow/issues/272, `SAMKNNClassifier` cannot use the `reset()` method. This seems to be due to the name mismatch between internal variable `self.max_wind_size` and input argument `max_window_size`.

`max_wind_size` does not appear to be referenced anywhere else in the library, so simply editing the name to match `max_window_size` in `SAMKNNClassifier.__init__()` may be all the change that is needed to fix this.

* 
* Rename `max_wind_size` to `max_window_size` on all three uses in `SAMKNNClassifier.__init__()`.
* Fixes # 272

---

#### Implementation
- [x ] Implementation is correct (it performs its intended function).
- [x ] Code is consistent with the framework.
- [x ] Code is properly documented.
- [x ] PR description covers ALL the changes performed.
- [x ] Files changed (update, add, delete) are in the PR's scope (no extra files are included).

#### Tests
- [x ] New functionality is tested.
- [ ] ALL tests pass with no errors.

### DIFF
--- a/src/skmultiflow/lazy/sam_knn.py
+++ b/src/skmultiflow/lazy/sam_knn.py
@@ -112,7 +112,7 @@ class SAMKNNClassifier(BaseSKMObject, ClassifierMixin):
         super().__init__()
         self.n_neighbors = n_neighbors
         self.weighting = weighting
-        self.max_wind_size = max_window_size
+        self.max_window_size = max_window_size
         self.ltm_size = ltm_size
         self.min_stm_size = min_stm_size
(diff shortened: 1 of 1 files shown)


ginop:
No checks were failed, but two were canceled. As far as I can tell, this was not caused by the PR and the edited code does not need further changes. What should I do to resolve this?

